PR #199, opened 2016-02-14
Iapp Parser refactor

pjbreaux:
#### What's this change do?
Changes the iapp parser to return a more JSON-like dictionary after parsing flat iapp template.

#### Any background context?
This change is preferred to allow better consumption of the iapp parser by f5 openstack heat plugins.

Test change then code change.


f5/common/iapp_parser.py
```diff
@@ -100,7 +105,7 @@ def get_template_name(self):
```
zancas: When is "`split_start[3][-1:]`" != "`u{`" ?

pjbreaux:
Addressed some of the comments that were given offline.

PR #203, opened 2016-02-17
_build_meta_data.. and _manage_.. --> activate_URI

zancas:
@pjbreaux 
Issues:
Fixes https://github.com/F5Networks/f5-common-python/issues/202

Problem: We used to have logic that all belonged in one function spread across
a function and a decorator, that was called in multiple places.

Analysis: We've simplified the code by moving that logic into a single function
and re-naming the function intelligently.

Tests: unit, and functional see changed unittest.


pjbreaux:
Everything looks swell here. Merging in.

PR #212, opened 2016-02-21
Subset of modules needed by LBaaSv2 and Heat projects

### LINKED ISSUE #197: Verify subset of f5-common modules needed by LBaaSv2 and Heat projects
Verify that f5-common-python modules can be used by other projects:

The following modules are needed by LBaaSv2 (f5-openstack-agent):

cm:
  cluster.py
  device.py

ltm:
  ~~monitor.py~~
  ~~pool.py~~
  ~~rule.py~~
  ~~snat.py~~
  ssl.py
  ~~virtual.py~~

net:
  arp.py
  interface.py
  l2gre.py
  route.py
  ~~selfip.py~~
  ~~vlan.py~~
  vxlan.py

sys:
  stat.py
  system.py

swormke:
@zancas @jlongstaf 

Fixes #197 

#### What's this change do?
We need to make sure that we support the objects that are needed by the various OpenStack projects that are in progress.  This change adds all of the objects that were being used via the REST API.

These are the same patterns used for the other objects.  It should be pretty straight forward.  I added some unit tests for things that didn't need to go to the actual device.

Also there is one new mixin to deal with things that don't have a named URI.

#### Any background context?
There are a few things that the old library looks like it is doing with the SOAP API.  We will need to figure out where these go and add them in using REST if possible.

#### Tests
* Flake8
* New functional and unit tests
* Old functional and unit tests

### DIFF
--- a/f5/bigip/mixins.py
+++ b/f5/bigip/mixins.py
@@ -108,3 +108,52 @@ class ExclusiveAttributesMixin(object):
                     [self.__dict__.pop(n, '') for n in new_set]
         # Now set the attribute
         super(ExclusiveAttributesMixin, self).__setattr__(key, value)
+
+
+class UnnamedResourceMixin(object):
+    '''This makes a resource object work if there is no name.
+
+    These objects do not support create or delete and are often found
+    as Resources that are under an organizing collection.  For example
+    the `mgmt/tm/sys/global-settings` is one of these and has a kind of
+    `tm:sys:global-settings:global-settingsstate` and the URI does not
+    match the kind.
+    '''
+    class UnsupportedMethod(Exception):
+        pass
+
+    def create(self, **kwargs):
+        '''Create is not supported for unnamed resources
+
+        :raises: UnsupportedOperation
+        '''
+        raise self.UnsupportedMethod(
+            "%s does not support the create method" % self.__class__.__name__
+        )
+
+    def delete(self, **kwargs):
+        '''Delete is not supported for unnamed resources
+
+        :raises: UnsupportedOperation
+        '''
+        raise self.UnsupportedMethod(
+            "%s does not support the delete method" % self.__class__.__name__
+        )
+
+    def load(self, **kwargs):
+        return self._load(**kwargs)
+
+    def _load(self, **kwargs):
+        '''Override _load because Unnamed Resources use their uri directly.
+
+        The Unnamed resources don't have URIs that match their kinds so
+        we need to use their URI directly instead of the container's URI
+        with name/partitions.
+        '''
+        self._check_load_parameters(**kwargs)
+        kwargs['uri_as_parts'] = True
+        read_session = self._meta_data['bigip']._meta_data['icr_session']
+        base_uri = self._meta_data['uri']
+        response = read_session.get(base_uri, **kwargs)
+        self._local_update(response.json())
+        return self
--- a/f5/bigip/sys/db.py
+++ b/f5/bigip/sys/db.py
@@ -32,16 +32,28 @@ class Db(Resource):
         self._meta_data['required_json_kind'] = 'tm:sys:db:dbstate'
 
     def create(self, **kwargs):
+        '''Create is not supported for db resources.
+
+        :raises: UnsupportedOperation
+        '''
         raise UnsupportedOperation(
             "DB resources doesn't support create, only load and refresh"
         )
 
     def update(self, **kwargs):
+        '''Update is not supported for db resources.
+
+        :raises: UnsupportedOperation
+        '''
         raise UnsupportedOperation(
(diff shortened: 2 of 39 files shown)


f5/bigip/mixins.py
```diff
@@ -108,3 +108,52 @@ def __setattr__(self, key, value):
...
+
+class UnnamedResourceMixin(object):
+    '''This makes a resource object work if there is no name.
+
+    These objects do not support create or delete and are often found
+    as Resources that are under an organizing collection.  For example
+    the `mgmt/tm/sys/global-settings` is one of these and has a kind of
+    `tm:sys:global-settings:global-settingsstate` and the URI does not
+    match the kind.
+    '''
```
zancas: There is no unique component/element/part of this resources URI-path.   We still say it's a Resource, because it's obviously NOT a Collection, _but_ it obviously can't support `create`, or `delete` since those operations are intrinsically tied to the unique parts of the URI (maybe we should add a check to delete! That says, if no `self._meta_data['uri']` then raise!).

If I Understand Correctly (IIUC) then I think we should also remove support for `load`, since it adds nothing that `refresh` doesn't supply...   right?!!?   Do we allow `load` on e.g. OrganizingCollections?

zancas: I just checked, we don't support `load` on anything but `Resource`.   I think this is correct, I think `load` is like `create` and `delete` in that it is only really intended to be used with server-resources that have unique elements in their URI-path.

f5/bigip/sys/db.py
```diff
@@ -32,16 +32,28 @@ def __init__(self, dbs):
         self._meta_data['required_json_kind'] = 'tm:sys:db:dbstate'
```
zancas: So is "`Db`" a class that should inherit from "`UnnamedResourceMixin`"?   It seems like it probably is.  I'm guessing it was defined before the mixin.

f5/bigip/sys/failover.py
```diff
@@ -0,0 +1,37 @@
...
+    To force the unit to standby call the `update()` method as follows:
+
+    ``f.update(command='run', standby=None, trafficGroup='mytrafficgroup')``
+    '''
+    def __init__(self, sys):
+        super(Failover, self).__init__(sys)
+        endpoint = self.__class__.__name__.lower()
+        self._meta_data['required_refresh_parameters'] = set()
+        self._meta_data['required_json_kind'] =\
+            'tm:sys:failover:failoverstats'
```
zancas: So, if my kind is: "`tm:SPAM:FOO:FOOstats`" am I automaticaly an "`UnnamedResourceMixin`"?  It'd be nice if that were the defining characteristic of "stats".   Also it'll make the Python-from-JSON jinja2 engine easier to augment.

f5/bigip/sys/performance.py
```diff
@@ -0,0 +1,55 @@
...
+from f5.bigip.resource import Resource
+from f5.bigip.resource import UnsupportedOperation
+
+
+class Performance(Collection):
+    def __init__(self, sys):
+        super(Performance, self).__init__(sys)
+        self._meta_data['allowed_lazy_attributes'] = [All_Stats]
+        self._meta_data['uri'] =\
+            self._meta_data['container']._meta_data['uri'] + "performance/"
```
zancas: I think we should rename this class to "`Performances`" and remove the explicit handling of "`self._meta_data['uri']`" from its constructor since that should be handled automatically in the call to "`super`".

f5/bigip/sys/performance.py
```diff
@@ -0,0 +1,55 @@
...
+    def __init__(self, performance):
+        super(All_Stats, self).__init__(performance)
+        self._meta_data['required_refresh_parameters'] = set()
+        self._meta_data['required_json_kind'] =\
+            'tm:sys:performance:all-stats:all-statsstats'
+        self._meta_data['uri'] =\
+            self._meta_data['container']._meta_data['uri'] + "all-stats/"
+
+    def update(self, **kwargs):
+        '''Update is not supported for statistics.
```
zancas: Maybe we should make a "`class StatsMixin(UnnamedResourceMixin):`" that overrides "`update`".

f5/bigip/sys/test/test_performance.py
```diff
@@ -0,0 +1,54 @@
...
+    fake_sys = mock.MagicMock()
+    return Performance(fake_sys)
+
+
+class TestPerformance(object):
+    def test_get_collection_raises(self):
+        perf = FakePerformance()
+        with pytest.raises(UnsupportedOperation):
+            perf.get_collection()
+
```
zancas: I think we should _usually_ have at least 1 `assert` statement per unittest.

PR #217, opened 2016-02-22
Node module needed to manage real servers #118

### LINKED ISSUE #118: Node module needed to manage real servers
Although this functionality could be put into the pool module (since real servers are often associated with pools), I believe it would be better implemented as its own module. It would allow creation, deletion, setup of health monitors etc...

On the BigIP the node lives in tmos.ltm.node

swormke:
@zancas
Issues:
Fixes #118

Problem:
Although this functionality could be put into the pool module (since real servers are often associated with pools), I believe it would be better implemented as its own module. It would allow creation, deletion, setup of health monitors etc...

On the BigIP the node lives in tmos.ltm.node

Analysis:
* Added Nodes and Node for LTM
* Added functional tests
* Moved ltm tests to a subdirectory

Tests:
* New CURDL and collection tests for Node/Nodes


zancas:
NOTE:  test_RUL fails when I run the functional test suite.   Did we really expect generation numbers to match?

PR #220, opened 2016-02-23
Template and Service resources in Applications should required partition

### LINKED ISSUE #200: Template and Service resources in Applications should required partition as creation parameter
The 'partition' should be a required_creation_parameter for the template and service resources.

swormke:
@pjbreaux 

Take a look at the code I added because I assume you also want `partition` for the `template.load()` so I added it as a required refresh parameter.  Maybe this isn't right?

Also added unit and functional tests for the new arg.

Issues:
Fixes #200

Problem:
Partition should be requred for template and services for iApps when they are created instead
of letting the system add whatever partition it wants.

Analysis:
* Added 'partition' to the `self._meta_data['required_creation_args']`
* Added 'partitoin' to the `self._meta_data['required_refresh_args']`
* Added unit and functional tests to cover the new parameters.

Tests:
* Ran all functional tests
* Ran all unit tests

### DIFF
--- a/f5/bigip/sys/application.py
+++ b/f5/bigip/sys/application.py
@@ -73,7 +73,9 @@ class Services(Collection):
 class Service(Resource):
     def __init__(self, service_s):
         super(Service, self).__init__(service_s)
-        self._meta_data['required_creation_parameters'].update(('template',))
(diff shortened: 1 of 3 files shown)


pjbreaux:
Looks good. Merging.

PR #221, opened 2016-02-23
Sys folder should require subPath as a creation parameter #194

### LINKED ISSUE #194: Sys folder should require subPath as a creation parameter
The subPath appears to be required when creating a sys folder on the BigIP. This parameter should be a 'required_creation_parameter' for that resource.

swormke:
@pjbreaux 
Issues:
Fixes #194

Problem:
The subPath appears to be required when creating a sys folder on the BigIP. This parameter should be a `required_creation_parameter` for that resource.

Analysis:
* Added the subPath as a required creation parameter
* Added a unit test to test if it is missing.

Tests:
* Ran folder unit and functional tests


pjbreaux:
Excellent. Looks good. Merging.

Issue #218, opened 2016-02-23
The template engine/library must produce an appropriate Python module for a given JSON blob


zancas:
Inputs:
* Template Library
* JSON blob
Outputs:
* Python code as strings:
    1.   import section
    2.   class section

zancas:
This is complete for "`OrganizingCollection`", "`Collection`", and "`Resource`".   I need to define "`Stats`" and "`URIPlaceHolder`" before I can add templates for them, so I think we can close this issue without those templates.

zancas:
I still need to clean up the code and add tests of the template gen code.

PR #222, opened 2016-02-24
Implement OC, Coll, Resource from temp generation.

zancas:
@pjbreaux 
Issues:
Fixes: https://github.com/F5Networks/f5-common-python/issues/218

Problem: There's a lot of boilerplate involved in creating new concrete
subclasses.  This engine provides a tool to automatically generate that
boilerplate.

Analysis: The engine requires templates and JSON-config data as inputs
it produces Python boilerplate as a string as output.

Tests: These will be added in other stories.

Where to start:

Pass "RESTWalker.py" to a python.   You should get a print out of a Resource based on the json in the "`device_configs`" directory and the "`Resource.tmpl`" in the "template_library" directory.


utils/template_library/Resource.tmpl
```diff
@@ -0,0 +1,8 @@
+
+
+class {{ klass }}(Resource):
+    def __init__(self, {{ container }}):
+        super({{ klass }}, self).__init__({{ container }})
+        self._meta_data['template_generated'] = True
+        self._meta_data['required_json_kind'] = {{ kind }}
+        self._meta_data['attribute_registry'] = {{ attr_reg_dict }}
```
pjbreaux: This is getting generated as follows:
```
class Pool(Resource):
    def __init__(self, Pools):
        super(Pool, self).__init__(Pools)
        self._meta_data['template_generated'] = True
        self._meta_data['required_json_kind'] = tm:ltm:pool:poolstate
        self._meta_data['attribute_registry'] =
```

The attribute_registry should be a dict.

utils/RESTWalker.py
```diff
@@ -0,0 +1,194 @@
...
+    def _format_org_collection(self, org_match, kind, raw_conf):
+        klass = org_match.group('OrgColl').capitalize()
+        OC_template = self.templates['OrganizingCollection']
+        import_dicts = self._build_import_dicts(raw_conf, klass)
+        config_dict = {'klass': klass,
+                       'kind': kind,
+                       'import_dicts': import_dicts}
+        OrgCollstr = OC_template.render(**config_dict)
+        imports_as_str = self.import_template.render(import_dicts=import_dicts)
+        imps_as_list = imports_as_str.split('\n')
```
pjbreaux: Try splitlines instead.

utils/RESTWalker.py
```diff
@@ -0,0 +1,194 @@
...
+        imps_as_list.append(
+            'from f5.bigip.resource import OrganizingCollection')
+        imps_as_list.sort()
+        imports = '\n'.join(imps_as_list)
```
pjbreaux: Is there a joinlines? Probably not.

utils/RESTWalker.py
```diff
@@ -0,0 +1,194 @@
...
+        else:
+            raise Kindless(raw_conf)
+        return python_as_string
+
+    def _klassify(self, raw_klass):
+        caps = '_'.join([x.capitalize() for x in raw_klass.split('-')])
+        dotless = '_'.join([x.capitalize() for x in caps.split('.')])
+        return dotless
+
+    def _build_import_dicts(self, raw_conf, klass):
```
pjbreaux: This is only used for OrganizingCollections. Perhaps rename?

utils/RESTWalker.py
```diff
@@ -0,0 +1,194 @@
...
+        imports = []
+        selfLinkstart = raw_conf[u"selfLink"].partition("?")[0]
```
pjbreaux: Does this fail if no '?' in link?

utils/RESTWalker.py
```diff
@@ -0,0 +1,194 @@
...
+        items = raw_conf[u"items"]
+        for item in items:
+            if item[u"reference"][u"link"].startswith(selfLinkstart):
+                tempuri = item[u"reference"][u"link"]
+                pre_qm = tempuri.partition("?")[0]
+                post_sl = pre_qm[len(selfLinkstart)+1:]
+                caps = self._klassify(post_sl)
+                imports.append({'OC': '.'+klass.lower(),
+                                'module': '.'+caps.lower(),
+                                'klass': caps})
```
pjbreaux: Should we log or print out when item[reference][link] does not match selfLinkstart?

utils/RESTWalker.py
```diff
@@ -0,0 +1,194 @@
...
+        return dotless
+
+    def _build_import_dicts(self, raw_conf, klass):
+        imports = []
+        selfLinkstart = raw_conf[u"selfLink"].partition("?")[0]
+        items = raw_conf[u"items"]
+        for item in items:
+            if item[u"reference"][u"link"].startswith(selfLinkstart):
+                tempuri = item[u"reference"][u"link"]
+                pre_qm = tempuri.partition("?")[0]
```
pjbreaux: pre_questionmark and post_selflink

utils/RESTWalker.py
```diff
@@ -0,0 +1,194 @@
...
+                post_sl = pre_qm[len(selfLinkstart)+1:]
+                caps = self._klassify(post_sl)
```
pjbreaux: Call it 'classified'?

utils/RESTWalker.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def process_config(self, config_name):
+        raw_conf = self.raw_configs[config_name]
+        if 'kind' in raw_conf:
+            python_as_string = self._process_config_with_kind(raw_conf)
+        else:
+            raise Kindless(raw_conf)
+        return python_as_string
+
+    def _klassify(self, raw_klass):
```
pjbreaux: Rename to something like 'handle_dashes_dots_and_case'.

utils/RESTWalker.py
```diff
@@ -0,0 +1,194 @@
...
+        template = self.templates['Resource']
+        python_as_string = template.render(container=container,
+                                           klass=raw_klass,
+                                           kind=kind)
+        return python_as_string
+
+    def _format_collection(self, kind, raw_conf):
+        raw_string = kind.split(':')[-1][:-len('collectionstate')]
+        raw_klass = self._klassify(raw_string)
+        if raw_klass.endswith('s'):
```
pjbreaux: Make its own function for readability and testability.

utils/RESTWalker.py
```diff
@@ -0,0 +1,194 @@
...
+            coll_klass = raw_klass + '_s'
+        else:
+            coll_klass = raw_klass + 's'
+        member_klass = raw_klass
+        collection_template = self.templates['Collection']
+        memberkind = kind.replace('collection', '')
+        config_dict = {'coll_klass': coll_klass,
+                       'member_klass': member_klass,
+                       'collection_container': kind.split(':')[1],
```
pjbreaux: What if there are subcollections?

utils/RESTWalker.py
```diff
@@ -0,0 +1,194 @@
...
+                       'collection_attr_reg_dict': {memberkind: member_klass}}
```
pjbreaux: For member_kind, what if you're a subcollection?

utils/RESTWalker.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def _format_stats(self, kind, raw_conf):
+        klass = kind.split(':')[-1][:-len('stats')]
+        print(klass)
+        template = self.templates['Stats']
+        print(template)
+        # XXX
+        # return python_as_string
+
+    def _process_config_with_kind(self, raw_conf):
```
pjbreaux: Identify why order matters in the docstring for your org_match etc...

utils/RESTWalker.py
```diff
@@ -0,0 +1,194 @@
...
+        imps_as_list.sort()
+        imports = '\n'.join(imps_as_list)
+        python_as_string = self.license_template.render()+imports+OrgCollstr
+        self.formatted_configs = {klass: {'Python_str': python_as_string,
+                                          'config_dict': config_dict,
+                                          'import_dicts': import_dicts}}
+        return python_as_string
+
+    def _format_resource(self, kind, raw_conf):
+        raw_string = kind.split(':')[-1][:-len('state')]
```
pjbreaux: This resource name is often the second element from the end? Is it always the case? And do you want to know if that's not the case?

utils/RESTWalker.py
```diff
@@ -0,0 +1,194 @@
...
+        raw_klass = self._klassify(raw_string)
+        if raw_klass.endswith('s'):
+            container = raw_klass+'_s'
+        else:
+            container = raw_klass+'s'
+        template = self.templates['Resource']
+        python_as_string = template.render(container=container,
```
pjbreaux: attr_reg_dict should be empty dict in output string.

utils/RESTWalker.py
```diff
@@ -0,0 +1,194 @@
...
+    return obj
+
+
+def dump_json(obj, pathname):
+    fp = io.open(pathname, 'wb')
+    json.dump(obj, fp, indent=1)
+    fp.close()
+
+
+class TemplateEngine(object):
```
pjbreaux: This is puntable, but I'd like to see docstrings for the reasons why this thing exists. And possible info on the edge cases.

utils/RESTWalker.py
```diff
@@ -0,0 +1,194 @@
...
+        os.path.walk(dirname, consumer, self)
+
+    def list_templates(self):
+        pp(self.templates.keys())
+
+    def list_raw_configs(self):
+        pp(self.raw_configs.keys())
+
+    def process_config(self, config_name):
+        raw_conf = self.raw_configs[config_name]
```
pjbreaux: If config_name not found (KeyError), tell them that it's a filename, and where it has to be located. All that jazz.

utils/RESTWalker.py
zancas: I propose we punt this:  https://github.com/F5Networks/f5-common-python/issues/223

PR #236, opened 2016-03-04
Feature.iapp parser double quote braces

pjbreaux:
@zancas 

#### What's this change do?
Adds functionality to ignore curly braces inside double quotes if those are present in an iapp template.

#### Any background context?
This stemmed from an issue of parsing an iapp template with a curly braces in a double quoted string like so: ```implementation { ... ... ... { } { hello test } hello quoted curly brace "}" ... }```

New test, then look at the code changes


zancas:
NOTE:   In tclsh (at least in some contexts) {foo} has the same semantics as "foo".

PR #241, opened 2016-03-05
add base exception

zancas:


swormke:
Should we call it `F5SDKError`?

A doc string for this would be nice too and an entry in docs/apidoc/f5.bigip.rst in the Resource Exceptions section.

zancas:
@pjbreaux Adding the docstring to the rst file shouldn't block.

pjbreaux:
Ship it!

Issue #233, opened 2016-03-03
Use refresh() in application.services.service _create() instead of load()


pjbreaux:
The load operation can fail after a create is done, because load uses kwargs and those kwargs may cause the BigIP get request to fail. We can populate the object with refresh instead.

zancas:
I don't understand the failure mode here.   Do we have a test demonstrating that load fails?   How can "loads" kwargs cause the bigip's get request to fail?

pjbreaux:
The sys.application.services.service overrides _create and _load. If kwargs are passed to _create, those are also given to _load (https://github.com/F5Networks/f5-common-python/blob/0.1/f5/bigip/sys/application.py#L128), and in the case of kwargs['trafficGroup'] being given to _load, it causes the get request there to fail. But instead of doing a _load in the _create of that resource, @swormke suggested doing a refresh().

zancas:
See PR #248

PR #248, opened 2016-03-08
Bugfix: The load operation can fail after a create is done, because load useskwargs and those kwargs may cause the BigIP get request to fail.

pjbreaux:
@zancas 

Issues: Issue #233

Problem: Load for a service will fail after create is called because all kwargs are passed from create to load

Analysis: Decided to make a resource meta_data property called 'disallowed_load_parameters' that will be stripped from kwargs passed into load if they are present.

Tests: Added the trafficGroup to one of the functional tests to ensure this does not fail on a real BigIP. Still testing at 100%


zancas:
LGTM

PR #255, opened 2016-03-09
Db objects now support update

### LINKED ISSUE #251: Add update() support to sys.db
bigip.sys.dbs.db does not support update(), but we have a use case where setting a value on sys.db is needed. Here's some sample code:

db = bigip.sys.dbs.db.load(name='iptunnel.configsync')
db.update(value='enable')

zancas:
@jlongstaf 
Issues:
Fixes #251

Problem: We had not previously implemented update for Db objects, but they are
supported.

Analysis: Implemented and tested.

Tests: sys/test_db::TestDb::test_update

### DIFF
--- a/f5/bigip/sys/db.py
+++ b/f5/bigip/sys/db.py
@@ -58,15 +58,6 @@ class Db(Resource):
             "DB resources doesn't support create, only load and refresh"
         )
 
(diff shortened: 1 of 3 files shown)


pjbreaux:
Ship it!

PR #291, opened 2016-03-18
Make requirements have SingleSourceofTruth

### LINKED ISSUE #290: Requirements are confused and out of date
We'll update using the logic explained here:

https://caremad.io/2013/07/setup-vs-requirement/

Basically (for a library) will have minimal need for "requirements.txt".   We'll use it only for test dependencies.

zancas:
@swormke
Issues:
Fixes #290

Problem: We had old requirements listed, and more than 1 source of truth.

Analysis: Using the pattern recommended here:

https://caremad.io/2013/07/setup-vs-requirement/

we've enforced a single source of truth, and removed obsolete requirements.

For general use of the f5-sdk library, add strings to the "install_requires"
parameter to the setup method in "setup.py".   For hacking on the codebase,
testing, etc. add packages and versions to the end of "requirements.txt".

NOTE:   The "-e ." line in the requirements.txt ensures that "pip install -r"
-ing that file will pick up the "install_requires" packages.

Tests: By hand verification of correct installation.


swormke:
What about the requirements.test.txt and requirements.docs.txt?

zancas:
/me waves hand:
"This is not the requirements.test.txt you're looking for."

zancas:
@swormke  manually tested against the PyCharm IDE to verify compatibility.

PR #305, opened 2016-03-24
PathElement abstract class for nonREST URI element

### LINKED ISSUE #304: Need to support RESTless URI path elements (e.g. 'shared/licensing')

zancas:
@pjbreaux
Issues:
Fixes #304

Problem: Some server provided path elements do not support HTTP operations

Analysis:  See for example the response to an HTTP GET against
'shared/licensing'

Tests: Current functional tests pass.


pjbreaux:
Ship it!

PR #327, opened 2016-03-30
Add hostname as a BigIP property #326

### LINKED ISSUE #326: Add hostname as a BigIP property
The `BigIP` object needs to have a property to retrieve the hostname of the device.  We store this in `_meta_data['hostname']` but that implies it is a private attributed of the object.  We need a more _public_ way to access it. 

@richbrowne @jlongstaf

swormke:
Issues:
Fixes #326

Problem:
The BigIP object needs to have a property to retrieve the hostname
of the device. We store this in _meta_data['hostname'] but that
implies it is a private attributed of the object. We need a more
public way to access it.

Analysis:
- Created a `@property` decorated method to return the hostname
  from `self._meta_data[]`

Tests:
- New unit test
- Flake8

### DIFF
--- a/f5/bigip/__init__.py
+++ b/f5/bigip/__init__.py
@@ -46,3 +46,7 @@ class BigIP(OrganizingCollection):
                            'device_name': None,
                            'local_ip': None,
                            'bigip': self}
+
+    @property
+    def hostname(self):
+        return self._meta_data['hostname']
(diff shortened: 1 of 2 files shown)


swormke:
@jlongstaf don't merge this yet, I thing I may have messed something up after looking at some functional tests.

swormke:
@jlongstaf Ok, it is fine, typo on my hostname :-)

PR #331, opened 2016-04-01
FDB needs to support a 11.5.0 api version only #314

### LINKED ISSUE #314: FDB needs to support a 11.5.0 api version only
Are you planning on adding FDB support here soon? It seems most of the work is done.

swormke:
@zancas @jlongstaf @richbrowne 

Issues:
Fixes #314

Problem:
The fdb API changed in v12.0, we want to make sure that this
call only calls API less than v12.0.

Analysis:
- Added fdb to the lazy attributes of net
- Added version support the the base bigip, set to '' by default
- Resources use their container's version

Test:
- Added unit tests
- Ran the functional tests
- Ran the unit tests
- Ran the flake8 tests

### DIFF
--- a/f5/bigip/__init__.py
+++ b/f5/bigip/__init__.py
@@ -34,19 +34,27 @@ class BigIP(OrganizingCollection):
         timeout = kwargs.pop('timeout', 30)
         allowed_lazy_attrs = kwargs.pop('allowed_lazy_attributes',
                                         allowed_lazy_attributes)
+        icontrol_version = kwargs.pop('icontrol_version', '')
         if kwargs:
             raise TypeError('Unexpected **kwargs: %r' % kwargs)
         # _meta_data variable values
         iCRS = iControlRESTSession(username, password, timeout=timeout)
         # define _meta_data
-        self._meta_data = {'allowed_lazy_attributes': allowed_lazy_attrs,
-                           'hostname': hostname,
-                           'uri': 'https://%s/mgmt/tm/' % hostname,
-                           'icr_session': iCRS,
-                           'device_name': None,
-                           'local_ip': None,
-                           'bigip': self}
+        self._meta_data = {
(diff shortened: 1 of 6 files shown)


f5/bigip/__init__.py
```diff
@@ -34,19 +34,27 @@ def __init__(self, hostname, username, password, **kwargs):
         timeout = kwargs.pop('timeout', 30)
         allowed_lazy_attrs = kwargs.pop('allowed_lazy_attributes',
                                         allowed_lazy_attributes)
+        icontrol_version = kwargs.pop('icontrol_version', '')
```
zancas, line 37: It seems like it would be more consistent with e.g. 'device_name' if we made this 'None'.

f5/bigip/net/fdb.py
```diff
@@ -64,6 +67,9 @@ def __init__(self, fdb):
         self._meta_data['attribute_registry'] =\
             {u'tm:net:fdb:tunnel:tunnelstate': Tunnel}
         self._meta_data['template_generated'] = True
+        # Setting this here to be explicit, even though it is set via
+        # the super call from its containing object.
+        self._meta_data['icontrol_version'] = '11.5.0'
```
zancas: If we made the class get its version from its container, and made its container pick up _its_ containers version, then we'd only need to set it explicitly in 'Fdbs'.

f5/bigip/net/fdb.py
```diff
@@ -75,3 +81,6 @@ def __init__(self, fdb):
         # self._meta_data['attribute_registry'] =\
         #     {u'tm:net:fdb:vlan:vlanstate': Vlan}
         self._meta_data['template_generated'] = True
+        # Setting this here to be explicit, even though it is set via
+        # the super call from its containing.
+        self._meta_data['icontrol_version'] = '11.5.0'
```
zancas: See note about getting it from its container.

PR #348, opened 2016-04-23
WIP: Adds support for connecting to BIG-IPs on off-numbered ports

### LINKED ISSUE #347: Add port option
I test BIG-IPs through a vagrant port forward, so the port is on an off number

ex
port 10443 forwarded to port 443 via vagrant

To support thing in the SDK I could make the hostname "localhost:10443", but supporting a port option that defaults to 443 sounds like it would make more sense.

I think an additional enhancement would be required in the icontrol rest library

caphrim007:
Issues:
Fixes #347

Problem:
The f5-sdk makes the assumption that all BIG-IPs are accessed on
port 443. While this may be true in most cases, there are cases
where the BIG-IP's web ui may be accessed through a port-forward
and therefore the f5-sdk would need to specify the forwarded port
instead of assuming the default.

Analysis:
I added a port argument to the BigIP constructor. By default this
value is the standard port that the web ui is accessed on; 443.

Whether the port is specified or not, the string formatting performed
for the self._meta_data['uri'] value will include the port value.
The new format of the 'uri' value will resemble the following

```
https://FakeHostName:443/mgmt/tm
```

If a port is specified, for instance 10443, it would then look like

```
https://FakeHostName:10443/mgmt/tm
```

This change allows BIG-IPs to be accessed on off-numbered ports

Tests:
- FakeBigIPWithPort
- test_meta_data_uri

### DIFF
--- a/f5/bigip/__init__.py
+++ b/f5/bigip/__init__.py
@@ -32,6 +32,7 @@ class BigIP(OrganizingCollection):
     """An interface to a single BIG-IP"""
     def __init__(self, hostname, username, password, **kwargs):
         timeout = kwargs.pop('timeout', 30)
+        port = kwargs.pop('port', 443)
         allowed_lazy_attrs = kwargs.pop('allowed_lazy_attributes',
                                         allowed_lazy_attributes)
         icontrol_version = kwargs.pop('icontrol_version', '')
@@ -43,7 +44,7 @@ class BigIP(OrganizingCollection):
         self._meta_data = {
             'allowed_lazy_attributes': allowed_lazy_attrs,
             'hostname': hostname,
-            'uri': 'https://%s/mgmt/tm/' % hostname,
+            'uri': 'https://%s:%s/mgmt/tm/' % (hostname, port),
             'icr_session': iCRS,
             'device_name': None,
             'local_ip': None,
--- a/f5/bigip/resource.py
+++ b/f5/bigip/resource.py
@@ -532,15 +532,15 @@ class Resource(ResourceBase):
         :param selfLinkuri: the server provided selfLink (contains localhost)
         :raises: URICreationCollision
         """
-        # hostname local alias
-        hostname = self._meta_data['bigip']._meta_data['hostname']
+        # netloc local alias
+        uri = urlparse.urlsplit(self._meta_data['bigip']._meta_data['uri'])
 
         # attrs local alias
         attribute_reg = self._meta_data.get('attribute_registry', {})
         attrs = attribute_reg.values()
 
         (scheme, domain, path, qarg, frag) = urlparse.urlsplit(selfLinkuri)
-        path_uri = urlparse.urlunsplit((scheme, hostname, path, '', ''))
+        path_uri = urlparse.urlunsplit((scheme, uri.netloc, path, '', ''))
         if not path_uri.endswith('/'):
             path_uri = path_uri + '/'
         qargs = urlparse.parse_qs(qarg)
--- a/f5/bigip/test/test___init__.py
+++ b/f5/bigip/test/test___init__.py
@@ -14,6 +14,7 @@
 
 import mock
 import pytest
+import urlparse
 
 from f5.bigip import BigIP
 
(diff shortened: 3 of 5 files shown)


caphrim007:
@swormke @zancas

f5/bigip/__init__.py
```diff
@@ -43,6 +44,7 @@ def __init__(self, hostname, username, password, **kwargs):
         self._meta_data = {
             'allowed_lazy_attributes': allowed_lazy_attrs,
             'hostname': hostname,
+            'port': port,
             'uri': 'https://%s/mgmt/tm/' % hostname,
```
swormke: Should the URI here be `'https://%s:%s/mgmt/tm/' % (hostname, port)`

swormke:
Can't believe we didn't have this from the beginning!  Thanks for adding.  I had one comment that I put inline about the _meta_data['uri'] in think that when we build that we need to also include the port right?

I'll let @zancas take a look too since he is closer to the code these days than I am.

f5/bigip/__init__.py
caphrim007: I originally thought that it should, but there is code further down in the library that uses the uri in some ways that I wasn't able to easily grab the port from....or something like that.

Lemme see if I can find my reasoning behind doing that. I think it was in the resource.py file.

caphrim007:
@swormke Now that I understand better how the SDK works, let me try again by tweaking the uri field

caphrim007:
@swormke @zancas I've made updates as @swormke suggested. It still seems to work on my end I think (if I havent hosed by virtualenv enough :-P) Have a look

f5/bigip/resource.py
```diff
@@ -532,15 +532,15 @@ def _activate_URI(self, selfLinkuri):
         :param selfLinkuri: the server provided selfLink (contains localhost)
         :raises: URICreationCollision
         """
-        # hostname local alias
-        hostname = self._meta_data['bigip']._meta_data['hostname']
+        # netloc local alias
+        uri = urlparse.urlsplit(self._meta_data['bigip']._meta_data['uri'])
```
zancas, line 536: Nice!  +1    So urlsplit is returning a NamedTuple?  I did not realize...   awesome.

f5/bigip/test/test___init__.py
```diff
@@ -42,3 +50,8 @@ def test___get__attr(FakeBigIP):
     with pytest.raises(AttributeError):
         FakeBigIP.this_is_not_a_real_attribute
     assert FakeBigIP.hostname == 'FakeHostName'
+
+
+def test_meta_data_uri(FakeBigIPWithPort):
```
zancas: This is slightly more specific than a just a test of "meta_data_uri" right, it's a test of the non-default port number, so let's rename it to something that includes the words "port", "number", and "default".

f5/bigip/__init__.py
```diff
@@ -30,7 +30,7 @@
 
 class BigIP(OrganizingCollection):
     """An interface to a single BIG-IP"""
-    def __init__(self, hostname, username, password, **kwargs):
+    def __init__(self, hostname, username, password, port=443, **kwargs):
```
zancas: We're using the keyword args default params pattern here, soo...

Instead of having `port=443` in the param list for the constructor, move it down into the "kwargs.pop"s.

So the whole param list should be:

`def __init__(self, hostname, username, password, **kwargs):`

and then...   see next line note...

f5/bigip/__init__.py
```diff
@@ -30,7 +30,7 @@
...
         timeout = kwargs.pop('timeout', 30)
         allowed_lazy_attrs = kwargs.pop('allowed_lazy_attributes',
                                         allowed_lazy_attributes)
```
zancas, line 37: Add this:

`port = kwargs.pop('port', 443)`

zancas, line 37: Then even though it's not explicitly about the port number add:

```
if kwargs:
    raise TypeError('Unexpected kwargs: %r' % kwargs)
```

zancas:
@caphrim007 
OK, I've added line notes with all my comments.

For the most part it looks great, and I'm excited to see that the return type of urlsplit is a NamedTuple...  cool!    Nice improvement.

I'd like us to continue to conform to a single pattern for default params.  We've decided on the `keyword-only` default params (From "Effective Python), so let's do the teeny refactor needed to move the `param=443` from a param to a kwarg key-value.   

See line notes for details --^

Once that's done I'll merge.

caphrim007:
@zancas changes incoming

PR #354, opened 2016-04-27
Adds System DNS API to the SDK

### LINKED ISSUE #353: Add System DNS configuration
There is no API for the System DNS settings. It should be added to allow the ability to easily change those settings

caphrim007:
@zancas @swormke 

Fixes #353 
#### What's this change do?

This patch adds the System DNS API

f5/bigip/sys/dns.py or f5/bigip/sys/test_dns.py

The System DNS settings can be changed in the UI as well as the API. In particular, if you have access to 12.0.0, you can access /mgmt/toc to navigate to the provided location and check that the settings there are modified accordingly.
#### Any background context?

This patch adds the DNS API so that you can set the system DNS settings
such as DNS search list and DNS cache options

### DIFF
--- a/f5/bigip/sys/dns.py
+++ b/f5/bigip/sys/dns.py
@@ -1,0 +1,46 @@
+# coding=utf-8
+#
+# Copyright 2016 F5 Networks Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+
+"""BIG-IP® system dns module
+
+REST URI
+    ``http://localhost/mgmt/tm/sys/dns``
+
+GUI Path
+    ``System --> Configuration --> Device --> DNS``
+
+REST Kind
+    ``tm:sys:dns:*``
+"""
+
+from f5.bigip.mixins import UnnamedResourceMixin
+from f5.bigip.resource import Resource
+
+
+class Dns(UnnamedResourceMixin, Resource):
+    """BIG-IP® system DNS unnamed resource
+
+        .. note::
+
+        This is an unnamed resource so it has not ~Partition~Name pattern
+        at the end of its URI.
+    """
+    def __init__(self, sys):
+        super(Dns, self).__init__(sys)
+        self._meta_data['required_load_parameters'] = set()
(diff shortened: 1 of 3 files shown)


f5/bigip/sys/dns.py
```diff
@@ -0,0 +1,48 @@
...
+        This is an unnamed resource so it has not ~Partition~Name pattern
+        at the end of its URI.
+    """
+    def __init__(self, sys):
+        super(Dns, self).__init__(sys)
+        endpoint = self.__class__.__name__.lower()
+        self._meta_data['required_load_parameters'] = set()
+        self._meta_data['required_json_kind'] = 'tm:sys:dns:dnsstate'
+        self._meta_data['uri'] = \
+            self._meta_data['container']._meta_data['uri'] + endpoint + '/'
```
zancas: This logic needs to be pushed into the `UnnamedResourceMixin`, it is (or should be) a defining feature of that Mixin.    It's not a blocker on this PR, but if we do not migrate it in this PR, then we should make it a blocking issue for future implementations of `UnnamedResourceMixin` concrete subclasses.

LOL @wojtek0806 pointed out that @swormke already implemented this:

https://github.com/F5Networks/f5-common-python/blame/0.1/f5/bigip/mixins.py#L164

OK, so we should use that method like this:

```
self._meta_data['uri'] = self.WHATEVER_THE_UNNAMED_RESOURCE_MIXIN_METHOD_IS
```

f5/bigip/sys/test/test_dns.py
```diff
@@ -0,0 +1,38 @@
...
+from f5.bigip.sys.dns import Dns
+
+
+@pytest.fixture
+def FakeDns():
+    fake_sys = mock.MagicMock()
+    return Dns(fake_sys)
+
+
+class TestDns(object):
```
zancas: We don't need a class scope here.   Let's remove `class TestDns` and "promote" its methods to module-level functions.

f5/bigip/sys/test/test_dns.py
```diff
@@ -0,0 +1,38 @@
...
+    def test_create_raises(self):
+        dns = FakeDns()
```
zancas: Once these tests have been promoted to module level functions, let's access `FakeDns` in the pytest fixture (NON-CANONICAL PYTHON_ fashion. 
Then let's also follow the convention of having tests end with an assertion in every reasonable instance.
  This is what this would look like:

```
def test_create_raises(FakeDns):
    with pytest.raises(UnnamedResourceMixin.UnsupportedMethod) as EIO:
        FakeDns.create()
    assert EIO.value.message == 'SOMETHING'
```

caphrim007:
@wojtek0806 you said you already had this coded? PR#?

wojtek0806:
@zancas  Just created one:  Feature.fix multiple bugs #356

caphrim007:
@zancas fixes incoming.

zancas:
@caphrim007 I'm still waiting on review-comment fixes...   correct?  Or did I get confused?

caphrim007:
@zancas, I think they were addressed in my recent push. If not, lemme know

PR #360, opened 2016-05-02
WIP: Add Transaction class to BIG-IP

### LINKED ISSUE #359: support for transactions
The REST API has a transaction feature that allows you to chunk together multiple operations and commit them atomically.

This feature is very useful for operations that touch many different parts of the BIG-IP, but if they were committed separately, might leave the BIG-IP in an inconsistent state.

caphrim007:
Issues:
Fixes #359

Problem:
The REST API supports a transaction feature that allows you to modify
many config objects and commit them all at once. if the modifications
fail, then the entire set is rolled back. This allows you to do atomic
operations on many objects without commiting something that would leave
the BIG-IP in an inconsistent state

Analysis:
The REST API supports transactions that, if they fail, will automatically
rollback all of the operations. I need to use this for some functional
use cases where individual modification of Configuration could result
in a BIG-IP being left in an inconsistent state.

I have modeled this feature off of Context Managers in python. You provide
an API object inside of a transaction, and then use that API object to change
whatever you want. When the block finishes, the entire thing is either committed
or rolled back

Tests:
None yet as this is a WIP and needs feedback from core developers.

### DIFF
--- a/f5/bigip/contexts.py
+++ b/f5/bigip/contexts.py
@@ -1,0 +1,79 @@
+# coding=utf-8
+#
+# Copyright 2014 F5 Networks Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+import copy
+import logging
+
+from f5.sdk_exception import F5SDKError
+
+
+class TransactionSubmitException(F5SDKError):
+    pass
+
+
+class TransactionContextManager(object):
+    def __init__(self, transaction):
+        self.transaction = transaction
+        self.bigip = transaction._meta_data['bigip']
+        self.icr = self.bigip._meta_data['icr_session']
+        self.original_headers = copy.deepcopy(self.icr.session.headers)
+
+    def __enter__(self):
+        """Begins a new transaction context
+
+        When a transaction begins, this method will automatically be called
+        to set up the transaction.
+
+        Transaction IDs are automatically retrieved for you and the
+        appropriate headers are set so that operations in the Transaction
+        Context will reference the transaction.
+
+        Headers are preserved so that after you exit the transaction, you will
+        be able to use the API object as you normally would.
+        """
+
+        self.transaction.create()
+        self.icr.session.headers.update({
+            'X-F5-REST-Coordination-Id': self.transaction.transId
+        })
+        return self.bigip
+
+    def __exit__(self, exc_type, exc_value, exc_tb):
+        """Commit a transaction upon Context Manager exit
+
+        Upon exit, the transaction will attempt to be committed. If the commit
+        fails, the transaction will automatically be rolled back by the server
+        performing the transaction.
+
+        :param exc_type: The type of exception raised
+        :param exc_value: Value of the exception raised
+        :param exc_tb: Traceback
+        NOTE: If the context exits without an exception, all three of the
+        parameters will be None
+        :returns: void
+        """
+        self.icr.session.headers = dict()
+        if exc_tb is None:
+            try:
+                self.transaction.update(state="VALIDATING")
+            except Exception as e:
+                logging.debug(e)
+                raise TransactionSubmitException(e)
+            finally:
+                self.icr.session.headers = self.original_headers
+        self.icr.session.headers = self.original_headers
--- a/f5/bigip/resource.py
+++ b/f5/bigip/resource.py
@@ -532,6 +532,7 @@ class Resource(ResourceBase):
         :param selfLinkuri: the server provided selfLink (contains localhost)
         :raises: URICreationCollision
         """
+
         # netloc local alias
         uri = urlparse.urlsplit(self._meta_data['bigip']._meta_data['uri'])
 
@@ -575,7 +576,8 @@ class Resource(ResourceBase):
         # Post-process the response
         self._local_update(response.json())
 
-        if self.kind != self._meta_data['required_json_kind']:
+        if self.kind != self._meta_data['required_json_kind'] \
+           and self.kind != "tm:transaction:commandsstate":
             error_message = "For instances of type '%r' the corresponding"\
                 " kind must be '%r' but creation returned JSON with kind: %r"\
                 % (self.__class__.__name__,
--- a/f5/bigip/transaction/__init__.py
+++ b/f5/bigip/transaction/__init__.py
@@ -1,0 +1,60 @@
+# coding=utf-8
+#
+# Copyright 2014 F5 Networks Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""BIG-IP® system dns module
+
+REST URI
+    ``http://localhost/mgmt/tm/transaction``
+
+REST Kind
+    ``tm:transaction*``
+"""
+
+from f5.bigip.resource import Collection
+from f5.bigip.resource import Resource
+
+
+class Transactions(Collection):
+    """This class is a context manager for iControl transactions.
+
+    Upon successful exit of the with statement, the transaction will be
+    submitted, otherwise it will be rolled back.
+
+    NOTE: This feature was added to BIGIP in version 11.0.0.
+
+    Example:
+    > bigip = BigIP(<args>)
+    > tx = bigip.transactions.transaction
+    > with TransactionContextManager(tx) as api:
+    >     api.net.pools.pool.create(name="foo")
+    >     api.sys.dbs.db.update(name="setup.run", value="false")
+    >     <perform actions inside a transaction>
+    >
+    > # transaction is committed when you exit the "with" statement.
+    """
+    def __init__(self, api):
+        super(Transactions, self).__init__(api)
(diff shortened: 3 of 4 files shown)


caphrim007:
@zancas @swormke This. I like the Transaction class, but I hate that I had to fudge the kinds variable in resource. Also, my code inside of Transaction looks kinda kludgy, but the idea is solid (Context Managers). Feedback/guidance pleeeeeeeaaaase.

caphrim007:
@zancas yeah it's a screwed up feature. I had to read DevCentral articles all day long to figure it out :-/

caphrim007:
@zancas which line?

f5/bigip/contexts.py
```diff
@@ -0,0 +1,61 @@
...
+        self.icr = self.bigip._meta_data['icr_session']
+
+    def __enter__(self):
+        """Begins a new transaction context
+
+        When a transaction begins, this method will automatically be called
+        to set up the transaction.
+
+        Transaction IDs are automatically retrieved for you and the
+        appropriate headers are set so that future operations will reference
```
zancas: We should probably say change "future operations" to something like: "operations in the Transaction Context", because it's not _all_ future operations, it's just the ones inside the "transaction".     This line of thinking makes me think we need to change the name of this class from:   "TransactionContext" to "TransactionContextManager".
What do you think?

f5/bigip/contexts.py
```diff
@@ -0,0 +1,61 @@
...
+        the transaction.
+
+        Headers are preserved so that after you exit the transaction, you will
+        be able to use the API object as you normally would.
```
zancas, line 47: Wait...   we didn't actually do this!

To make the above docstring line true we need to have something like:

``` Python
import copy
self.original_headers = copy.deepcopy(self.icr.session.headers)
```

Then we need to copy that back in `__exit__`...   unless an empty dict is better?

f5/bigip/contexts.py
```diff
@@ -0,0 +1,61 @@
...
+        parameters will be None
+        :returns: bool -- False if an exception occurred. This instructs
+        python to process the raised exception as it would any other
+        exception. True would cause it to suppress the exception from
+        propagating.
+        """
+        self.icr.session.headers = dict()
+        if exc_tb is None:
+            self.transaction.update(state="VALIDATING")
+        else:
```
zancas: We can remove this `else: raise`.

f5/bigip/contexts.py
caphrim007, line 47: i think copy them back is correct behavior because it sets the object back to how it was before all the transaction stuff happened.

f5/bigip/transaction/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+
+    Upon successful exit of the with statement, the transaction will be
+    submitted, otherwise it will be rolled back.
+
+    NOTE: This feature was added to BIGIP in version 11.0.0.
+
+    Example:
+    > bigip = BigIP(<args>)
+    > tx = bigip.transactions
+    > with Transaction(tx) as api:
```
zancas: This'll change to `TransactionContextManager(tx)`

f5/bigip/transaction/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+    """This class is a context manager for iControl transactions.
+
+    Upon successful exit of the with statement, the transaction will be
+    submitted, otherwise it will be rolled back.
+
+    NOTE: This feature was added to BIGIP in version 11.0.0.
+
+    Example:
+    > bigip = BigIP(<args>)
+    > tx = bigip.transactions
```
zancas: This'll change to `tx = bigip.transactions.transaction`

f5/bigip/transaction/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+            {'tm:transactionstate': Transaction}
+
+
+class Transaction(Resource):
+    def __init__(self, transactions):
+        super(Transaction, self).__init__(transactions)
+        self._meta_data['required_json_kind'] = 'tm:transactionstate'
+
+    def _create(self, **kwargs):
+        """wrapped by `create` override that in subclasses to customize"""
```
zancas: Are these just copies of `_create` and `_update` from `Resource`?   If yes, then remove them, they're inherited.

caphrim007: no. the _create removes the check for required params because the transaction has no required params and it was originally failing because name= was not specified. It also makes the _create_uri include the endpoint because stock _create does not do this, and it sets the json= to {} because when specifying json={} to the create function, it treats that as a kwarg and puts it in the request_params instead of json argument.

the _update method removes all of the request_params processing because the transaction api errors out if you send anything with a PUT.

zancas:
OK, thanks for the explanation.

zancas:
I prefer that we modify `self._meta_data['required_creation_parameters']` by removing `name` from it, rather than removing the check of `required_creation_params`, does that make sense?

zancas:
Double check the actual endpoint by inserting:

`print(self._meta_data['container']._meta_data['uri'])`

are you sure that there're the correct number of "`transaction`"s at the end of uri (shouldn't it be 1?)?

And please paste the result here, so I'm sure we're examining the same observables!  Thanks.

zancas:
Finally try to call it without "json=", just pass an empty dict.    To a version that has the default `_create`.

zancas:
Lemme see what that implies about the client code (TransactionContextManager)....

f5/bigip/contexts.py
```diff
@@ -0,0 +1,61 @@
...
+        to set up the transaction.
+
+        Transaction IDs are automatically retrieved for you and the
+        appropriate headers are set so that future operations will reference
+        the transaction.
+
+        Headers are preserved so that after you exit the transaction, you will
+        be able to use the API object as you normally would.
+        """
+        self.transaction.create(json={})
```
zancas: Let's try changing this to:

`self.transaction.create({})`

zancas:
I think if we change the caller like this:

https://github.com/F5Networks/f5-common-python/pull/360/files#r62067830

then we won't need to worry about our "{}" getting sucked into the wrong param.

f5/bigip/contexts.py
caphrim007: I get this when i do that
(ansible-2.0)SEA-ML-RUPP1:f5-common-python trupp$ python ok.py 
Traceback (most recent call last):
  File "ok.py", line 12, in <module>
    with TransactionContextManager(tx) as api:
  File "/Users/trupp/src/trupp/f5-common-python/f5/bigip/contexts.py", line 45, in **enter**
    self.transaction.create(empty)
TypeError: create() takes exactly 1 argument (2 given)

f5/bigip/contexts.py
```diff
@@ -0,0 +1,66 @@
...
+
+import copy
+
+
+class TransactionContextManager(object):
+    def __init__(self, transaction):
+        self.transaction = transaction
+        self.bigip = transaction._meta_data['bigip']
+        self.icr = self.bigip._meta_data['icr_session']
+        self.original_headers = None
```
zancas: I think this should be:

`self.original_headers = copy.deepcopy(self.icr.session.headers)` not `None`.

f5/bigip/contexts.py
```diff
@@ -0,0 +1,66 @@
...
+
+        Transaction IDs are automatically retrieved for you and the
+        appropriate headers are set so that operations in the Transaction
+        Context will reference the transaction.
+
+        Headers are preserved so that after you exit the transaction, you will
+        be able to use the API object as you normally would.
+        """
+
+        self.original_headers = copy.deepcopy(self.icr.session.headers)
```
zancas: Let's do this copy.deepcopy step above in the `TransactionContextManager` `__init__`.

We'll use `__enter__` to create and customize specific transactions.   We'll use `__init__` to setup context.

f5/bigip/contexts.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def __exit__(self, exc_type, exc_value, exc_tb):
+        """Commit a transaction upon Context Manager exit
+
+        :param exc_type: The type of exception raised
+        :param exc_value: Value of the exception raised
+        :param exc_tb: Traceback
+        NOTE: If the context exits without an exception, all three of the
+        parameters will be None
+        :returns: bool -- False if an exception occurred. This instructs
```
zancas: If `__exit__` returns anything but True, that means exceptions were not handled within the `with` block.

f5/bigip/contexts.py
```diff
@@ -0,0 +1,66 @@
...
+        python to process the raised exception as it would any other
+        exception. True would cause it to suppress the exception from
+        propagating.
```
zancas: This docstring recaps too much of the definition of `__exit__` lets make it just describe what _this_ `__exit__` is doing.

f5/bigip/contexts.py
```diff
@@ -0,0 +1,66 @@
...
+        """
+        self.icr.session.headers = dict()
+        if exc_tb is None:
+            self.transaction.update(state="VALIDATING")
+            self.icr.session.headers = self.original_headers
```
zancas: let's move the assignment of `self.icr.session.headers` to `self.original_headers` to after the `if` condition so it always occurs, without regard to the exceptions status.

And let's think about wrapping the `self.transaction.update` in a `try`-`except` with a finally clause that also sets the headers back to the originals.

f5/bigip/resource.py
```diff
@@ -575,7 +576,11 @@ def _create(self, **kwargs):
         # Post-process the response
         self._local_update(response.json())
 
-        if self.kind != self._meta_data['required_json_kind']:
+        kinds = []
+        kinds.append(self._meta_data['required_json_kind'])
+        kinds.append("tm:transaction:commandsstate")
```
zancas: Now that we've refactored the `TransactionContextManager` do you still get the bugs that motivated you to add this code?

caphrim007: no
f5.bigip.resource.KindTypeMismatch: For instances of type ''Pool'' the corresponding kind must be ''tm:ltm:pool:poolstate'' but creation returned JSON with kind: u'tm:transaction:commandsstate'

zancas: You said "no" but I think you meant "yes" still get the bug...

f5/bigip/contexts.py
```diff
@@ -0,0 +1,80 @@
...
+        parameters will be None
+        :returns: void
+        """
+        self.icr.session.headers = dict()
+        if exc_tb is None:
+            try:
+                self.transaction.update(state="VALIDATING")
+            except Exception as e:
+                self.icr.session.headers = self.original_headers
+                logging.log.debug(e)
```
zancas: My bad, this syntax is wrong delete:   '.log'.   It should be:
`logging.debug(e)`

PR #376, opened 2016-05-05
Adds the auth/password-policy api to the skd

### LINKED ISSUE #374: add the auth password-policy to the sdk
currently the sdk does not support any of the auth endpoints in the REST API. This api is needed to change the password policy settings on a bigip.

caphrim007:
Issues:
Fixes #374

Problem:
The issue is that the password_policy endpoint is not available in
the sdk

Analysis:
This change adds the password-policy api to the sdk. This change was
needed to allow access to these settings so that they can be updated

Tests:
- f5/bigip/auth/test/test_password_policy.py
- test/functional/auth/test_password_policy.py

### DIFF
--- a/f5/bigip/auth/__init__.py
+++ b/f5/bigip/auth/__init__.py
@@ -1,0 +1,40 @@
+# coding=utf-8
+#
+# Copyright 2015-2016 F5 Networks Inc.
+#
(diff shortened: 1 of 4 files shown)


caphrim007:
@zancas

Issue #367, opened 2016-05-04
Evaluate probable reduntant code inside mixin.py


wojtek0806:
Evaluate the UnnamedResourceMixin code and remove redundant code for load method to improve readability.

PR #380, opened 2016-05-06
Unnamedresource.fixes command fixes

### LINKED ISSUE #355: Verify Unnamed Resorce Classes Coding
The unnamed resources classes contain redundant code for self._meta_data['uri'] , they should be utilizing super class method instead. This is in line with comments made on #354

wojtek0806:
@zancas 
Fixes #355 #367 

What's this change do?

Removed redundant code from classes inheriting from unnamed resource. Specifically now child classes utilize _get_meta_data_uri methods to populate self._meta_data['uri']

view changes below

Any background context?

Since Unnamedresourcemixin super contained the method _get_meta_data_uri for populating self._meta_data['uri']. This was required change to improve readability.


caphrim007:
lgtm when Travis build failures are addressed

wojtek0806:
Yea I spotted what's wrong....I will fix in the morning

zancas:
Isn't `_get_meta_data_uri_hyphens` just like `_get_meta_data_uri`...  with extra functionality?

If it is then `_get_meta_data_uri_hyphens` logic should be rolled into `_get_meta_data_uri`.

We should just have 1 method, not 2.

wojtek0806:
It might be possible but as we discussed having that separate it is easier to control the outcome. Right now we are assuming that no endpoint will have _ in their name, this might change. We can do this either way, I am just careful and trying to anticipate ;)

caphrim007:
@wojtek0806 if underscores were added, then you would also see tmsh accepting underscores. This is unlikely to happen imho because they already use hyphens for everything and it would be a divergence from that standard

wojtek0806:
@caphrim007 I am not a fan of introducing problems (however unlikely they may be) where they can be avoided. I have been in support for a while now and that taught me to expect the unexpected, and i would suggest keeping that second method for clarity.

wojtek0806:
@zancas 
Rebased given the last major change.

zancas:
This code is incorporate via another PR.

PR #393, opened 2016-05-10
Add port support to conftest

### LINKED ISSUE #392: unable to specify port in py.test
The port option was not added as a supported option to py.test tests so functional tests on off-numbered bigip ports will fail

caphrim007:
Issues:
Fixes #392

Problem:
You are unable to specify the port when running py.test so you are unable
to test bigips on off-numbered ports

Analysis:
This patch adds an option for specifying the port value and defaults
to the standard 443 port

Tests:
none

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -24,6 +24,8 @@ def pytest_addoption(parser):
                      default="admin")
     parser.addoption("--password", action="store", help="BIG-IP REST password",
                      default="admin")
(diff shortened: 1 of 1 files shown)


caphrim007:
@zancas @pjbreaux

zancas:
FYI, we're moving away from option passing for test...

PR #394, opened 2016-05-10
Sync to group in cm failing to sync

### LINKED ISSUE #396: Add API support for user creation
The API supports user creation via the /mgmt/tm/auth/user endpoint. this should be supported in the sdk

pjbreaux:
@zancas 

Issues:
Fixes #396

Problem:
This is failing silently to sync the BigIP to group.

Analysis:
I have implemented a working example of syncing to and from group, yet
the caveat is that there is no good way to determine whether a group (or
device) has been synced. This is as far as I know. I perused each uri
path for the device_group and for the device itself, and came up with
nothing. Nevertheless, the tests pass.

Tests: Ran all requisite tests for cm (unit and functional)


wojtek0806:
@pjbreaux 
The 'f5/bigip/tm/cm/**init**.py' does not utilize command mixin, I have did a change and tested this against my v12, with success, so something like this should work:

```
 class Cm(OrganizingCollection, CommandExecutionMixin):
     """BIG-IP® Cluster Organizing Collection."""
     def __init__(self, tm):
         super(Cm, self).__init__(tm)
         self._meta_data['allowed_lazy_attributes'] = [
             Devices, Device_Groups, Traffic_Groups, Sync_Status, Add_To_Trust,
             Remove_From_Trust
         ]
     def exec_cmd(self, command, **kwargs):
         cmds = ['run']
         if command not in cmds:
             error_message = "The command value {0} does not exist" \
                             "Valid commands are {1}".format(command, cmds)
             raise InvalidCommand(error_message)
         return self._exec_cmd(command, **kwargs)

```

So the execution syntax would be for example something like this:

```
b = bigip.BigIP('ltm5' , 'admin', 'admin')
b.cm.exec_cmd('run', utilCmdArgs='config-sync from-group datasync-device-bigip.resttest.lab.net-dg')
```

pjbreaux:
Okay @zancas, this is ready for review.

PR #398, opened 2016-05-10
Feature.clustering

pjbreaux:
@zancas 
#### What's this change do?

Implements the first bare features of BigIP clustering and scaling.
#### Any background context?

There is a need to cluster multiple BigIP devices (>=2 to <=8) with two basic cluster types ('sync-only', 'sync-failover'). This is the first basic features of that functionality. There are a few additional structures that have been added here. One important one is the f5/cluster directory, which holds the handlers for managing a cluster, device_group, and trusted_peers. These modules are simply wrappers around BigIP resource command (cm.device_groups.device_group) to control the order and timing of clustering.

Start at the single functional test for 'sync-failover' and then move on to the f5/cluster/cluster_manager.py. After that, you can look at the trusted_peer_manager.py and the device_group_manager.py.

I hate using 'manager' here, but it best describes what we're trying to accomplish. To run the functional tests, you'll need to create at least three BigIPs. Come talk to me!


pjbreaux:
Will squash my commits tomorrow to be it more palatable.

pjbreaux:
Reviewing my commits and reopening with a bit better commit history.

Issue #349, opened 2016-04-25
Add LTM profile creation


wojtek0806:
As for now the SDK does not allow for creation of profiles. SDK needs to provide the ability to create profiles for LTM.

zancas:
There's still work to be done for profiles beside client-ssl.

zancas:
@wojtek0806 I pushed latest to this branch:  https://github.com/zancas/f5-common-python/tree/profiles_rebased_changed

   I get the following error when I try to test:

```
=================================== FAILURES ===================================
__________________________ test_ssl_profile_creation ___________________________

cleaner = None
symbols = <pytest_symbols.models.Symbols object at 0x7fdd002a6310>
tmpdir = local('/tmp/pytest-of-root/pytest-0/test_ssl_profile_creation0')

    def test_ssl_profile_creation(cleaner, symbols, tmpdir):
        testdatadir = tmpdir.mkdir('testdir')
        FAKEKEYFILENAME = 'fakekey.key'
        CERTFILENAME = 'NEWTESTCLIENTPROFILENAME.crt'
        certpath = testdatadir.join(CERTFILENAME)
        certpath.write(FAKE_CERT)
        keypath = testdatadir.join(FAKEKEYFILENAME)
        keypath.write(FAKE_KEY)
        mr = ManagementRoot(symbols.bigip_netloc,
                            symbols.bigip_username,
                            symbols.bigip_password,
                            port=symbols.bigip_port)
        uploader = mr.shared.file_transfer.uploads
        cert_registrar = mr.tm.sys.crypto.certs
        pp(cert_registrar.raw)
        key_registrar = mr.tm.sys.crypto.keys
        ssl_client_profile = mr.tm.ltm.profile.client_ssls.client_ssl
        uploader.upload_file(str(certpath))
        uploader.upload_file(str(keypath))
        cert_registrar.install_cert(CERTFILENAME)
        pp([cert.raw for cert in cert_registrar.get_collection()])
        key_registrar.install_key(CERTFILENAME, FAKEKEYFILENAME)
        pp([key.raw for key in key_registrar.get_collection()])
        ssl_client_profile.create(name='FOOOO',
                                  certname='/Common/NEWTESTCLIENTPROFILENAME.crt',
>                                 keyname='/Common/NEWTESTCLIENTPROFILENAME.key')

test/functional/ssl_profile_creation/test_ssl_profile_creation.py:93:
[...]

PR #408, opened 2016-05-12
Dnsresolver.patch

### LINKED ISSUE #384: Add DNS resolver configuration API
Currently there is no way of having DNS resolver API under /mgmt/tm/net. This needs to be added

wojtek0806:
Fixes #384 

What's this change do?
Implements API for DNS resolver configuration

Any background context?
DNS resolver was not configurable via the SDK, ability to configure the DNS resolver other than adding functionality, is a requirement for some LTM profiles, and lack of implementation would block #349 from implementation.

tm/net/dns_resolver.py
test/functional/tm/net/test_dns_resolver.py


zancas:
Merged into a different issue.

PR #413, opened 2016-05-12
Make MetaDataURIMixin for composeability

### LINKED ISSUE #412: Handle hyphenation in 1 way using a special purpose mixin

zancas:
Issues:
Fixes #412

Problem: Converting from Python class names to device uris can require,
among other things, replacing '_' with '-'.  The new mixin is the place for
classname-to-URI manipulation.

Analysis: created MetaDataURIMixin

Tests: all current tests pass


pjbreaux:
Shipping!

Issue #341, opened 2016-04-10
Add better testing for device trust feature


wojtek0806:
Find a way to test the results of setting up and tearing down device trust. With the lack of support for trustDomain setup via SDK, utilizing sync status resource under CM module would be a good option for testing the results.

Issue #49, opened 2015-12-09
Refactor ltm/ssl.py


zancas:

zancas:
Hmmm....   @swormke  this seems like something we'll need sooner, than later.   I think I'll try to wrap it up before the release.

zancas:
I'm implementing a different set of resources than those listed above because of:

https://devcentral.f5.com/articles/demystifying-icontrol-rest-part-5-transferring-files?adbsc=social_default55359446&adbid=666661582995255298&adbpl=tw&adbpr=8232052

wojtek0806:
I am implementing profiles in the #349 why is the work being duplicated again?

wojtek0806:
@zancas Sorry for reopening this but I looked at the code and I disagree with the way you have implemented it. You do not need to upload certificate to be able to create LTM Client SSL profile. 

The certs themselves are managed under System --> File Management --> Ssl certificate List. 
Proper way to deploy this would be to utilize file upload class to upload the certificate file. Then you can attach it to the profile. This is how it is done via GUI and this is how it is done via TMSH

wojtek0806:
Issue Review: Partially duplicate, will be addressed by #349

zancas:
I did upload and install cert and key via the SDK.    See classes in `share` and `cm`.

PR #422, opened 2016-05-17
Ltm.profiles

### LINKED ISSUE #349: Add LTM profile creation

wojtek0806:
Fixes #349, #341

This update introduces profile submodule to LTM, allowing to create LTM classes.

Problem: SDK did not have the APIs to create LTM profiles, this change has addressed that

Analysis: Introduced new profile.py sub-module which incorporated majority of LTM profiles.

CAVEAT: Ramcache, and wa-cache profiles have placeholder classes and are not implemented due to some issues I was having with matching tmsh syntax to REST. This needs further investigation with Product Development

Tests:
Unit Tests
Flake8
Functional Tests

### DIFF
--- a/f5/bigip/tm/ltm/__init__.py
+++ b/f5/bigip/tm/ltm/__init__.py
@@ -33,7 +33,6 @@ from f5.bigip.tm.ltm.monitor import Monitor
 from f5.bigip.tm.ltm.nat import Nats
 from f5.bigip.tm.ltm.node import Nodes
 from f5.bigip.tm.ltm.persistence import Persistences
-from f5.bigip.tm.ltm.profile import Profile
 from f5.bigip.tm.ltm.policy import Policys
 from f5.bigip.tm.ltm.pool import Pools
 from f5.bigip.tm.ltm.profile import Profile
--- a/f5/bigip/tm/ltm/__init__.py
+++ b/f5/bigip/tm/ltm/__init__.py
@@ -53,7 +53,7 @@ class Ltm(OrganizingCollection):
             Nats,
             Nodes,
             Persistences,
-            Profiles,
+            Profile,
             Policys,
             Pools,
             Rules,
--- a/f5/bigip/tm/ltm/__init__.py
+++ b/f5/bigip/tm/ltm/__init__.py
@@ -33,9 +33,9 @@ from f5.bigip.tm.ltm.monitor import Monitor
 from f5.bigip.tm.ltm.nat import Nats
 from f5.bigip.tm.ltm.node import Nodes
 from f5.bigip.tm.ltm.persistence import Persistences
-from f5.bigip.tm.ltm.profile import Profile
 from f5.bigip.tm.ltm.policy import Policys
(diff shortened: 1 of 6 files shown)


wojtek0806:
@zancas 
Please review as this update will overwrite your profile.py, which has been merged yesterday. Like i mentioned in the closed ticket my work has been again duplicated

wojtek0806:
@zancas There is a problem in your change, You have removed 
''''
self._meta_data['allowed_lazy_attributes']
''''
Which breaks this badly. Entire list self._meta_data['allowed_lazy_attributes'] has been removed only containing Client_Ssls entry. Please fix

wojtek0806:
Also as far as I can see each organizing collection contains the allowed_lazy_attributes list.

wojtek0806:
@zancas 

I have looked at your new testing for client SSL, and reworked your coding, it should be backwards compatible with what you did and at the same time not break the convention we are using in the SDK. Your old solution was deviating from it and that would cause unnecessary problems. Also your usage of 'key', 'cert' the way you had is depricated, and utilizing 'certKeyChain' array is preffered, which is what the below addresses too. I cannot seem to run your test for client_ssl so can you please verify the below locally, and if all OK i will push this to my fork and we can start merging:

```

class Client_Ssl(Resource):
    def __init__(self, client_ssls):
        super(Client_Ssl, self).__init__(client_ssls)
        self._meta_data['required_json_kind'] =\
            'tm:ltm:profile:client-ssl:client-sslstate'

    def create(self, **kwargs):
        """Allows creation of SSL profile with just key/cert names
        otherwise it will just require 'name' kwarg to be provided
        """
        if 'certname' and 'keyname' in kwargs:
            kwargs['name'] = kwargs['certname'][:-4]
            kwargs['certKeyChain'] =\
                [{'cert': kwargs['certname'], 'key': kwargs['keyname']}]

        self._create(**kwargs)
        return self

```

wojtek0806:
This addresses #49

wojtek0806:
My mistake i reworked the code, forgot about the kwargs removal :)

```
        if 'certname' and 'keyname' in kwargs:
            certname = kwargs.pop('certname')
            keyname = kwargs.pop('keyname')
            kwargs['name'] = certname[:-4]
            kwargs['certKeyChain'] =\
                [{'cert': certname, 'key': keyname}]
```

zancas:
'`allowed_lazy_attributes`' is a different variable than '`attribute_registry`' it would be a nice enhancement to make those two names more distinct.

wojtek0806:
allowed_lazy_attributes is a part of organizing collections in the code, and I have checked my coding and you have removed:

self._meta_data['allowed_lazy_attributes']

from profiles class, which disabled my profile classes :)

zancas:
Oops..

you're right I'll add that back.

wojtek0806:
closing this PR changes will be pushed via rebased branch

PR #424, opened 2016-05-17
Moves Transactions to underneath the ManagementRoot class

### LINKED ISSUE #423: Move Transactions into ManagementRoot
Transactions are a sub-api of ManagementRoot insofar as their api endpoint is /mgmt/tm/transaction. They should be moved underneath that API while preserving the Transaction class in the BigIP class.

caphrim007:
Issues:
Fixes #423

Problem:
The Transaction class is currently outside of the ManagementRoot
Class. Since it is a sub-api of that API, it needs to be moved
to underneath that API

Analysis:
This change just moves the Transaction API to underneath the
MR API.

Tests:
none


caphrim007:
@zancas

caphrim007:
test

PR #426, opened 2016-05-17
Add the software update API endpoint

### LINKED ISSUE #425: Add software update api
The API for specifying Software update checking is not in the SDK. Add it.

caphrim007:
Issues:
Fixes #425

Problem:
The software/update api endpoint is not available in the SDK

Analysis:
This patch adds the api and associated tests

Tests:
- tests/functional/tm/sys/software/test_update.py
- f5/bigip/tm/sys/software/test/test_update.py

### DIFF
--- a/f5/bigip/tm/sys/software/__init__.py
+++ b/f5/bigip/tm/sys/software/__init__.py
@@ -1,0 +1,44 @@
+# coding=utf-8
+#
+# Copyright 2015 F5 Networks Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""BIG-IP® System Software module
+
+REST URI
+    ``http://localhost/mgmt/tm/sys/software``
+
+GUI Path
+    ``System``
+
+REST Kind
+    ``tm:sys:software:*``
+"""
+
+from f5.bigip.resource import OrganizingCollection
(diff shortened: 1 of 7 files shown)


caphrim007:
test jenkins

caphrim007:
foo

f5/bigip/tm/sys/software/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+    ``tm:sys:software:*``
+"""
+
+from f5.bigip.resource import LazyAttributeMixin
+from f5.bigip.resource import PathElement
+from f5.bigip.resource import ToDictMixin
+from f5.bigip.tm.sys.software.update import Update
+
+
+class Software(PathElement, LazyAttributeMixin, ToDictMixin):
```
zancas: Why not subclass `OrganizingCollection`?

Also...   those mixins aren't defined in `resource.py`.

caphrim007: Ok, I will fix the mixins. So the problem with the OrganizingCollection is that it inherits ResourceBase which itself provides an update() method.

So when you use this API in practice like this

tm.sys.software.update.load()

you get an AttributeError because update is a function. :-/

f5/bigip/tm/sys/software/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+    ``tm:sys:software:*``
+"""
+
+from f5.bigip.mixins import LazyAttributeMixin
+from f5.bigip.mixins import ToDictMixin
+from f5.bigip.resource import PathElement
+from f5.bigip.tm.sys.software.update import Update
+
+
+class Software(PathElement, LazyAttributeMixin, ToDictMixin):
```
zancas: I don't understand why `Software` doesn't "just" inherit from `OrganizingCollection`?

caphrim007: Because update in the use of this class is not a function and organizing collection inherits from which defines an update function

caphrim007: Inherits from resourcebase

zancas:
Nice!   `return super(Foo, self).SPAM(HAM)` ....

PR #428, opened 2016-05-18
Includes #422 pull request changes and comment and rebase

### LINKED ISSUE #349: Add LTM profile creation

wojtek0806:
Fixes #349, #341, #49 , #422

This update introduces profile submodule to LTM, allowing to create LTM classes.

Problem: SDK did not have the APIs to create LTM profiles, this change has addressed that

Analysis: Introduced new profile.py sub-module which incorporated majority of LTM profiles.

CAVEAT: Ramcache, and wa-cache profiles have placeholder classes and are not implemented due to some issues I was having with matching tmsh syntax to REST. This needs further investigation with Product Development

Tests:
Unit Tests
Flake8
Functional Tests

### DIFF
--- a/f5/bigip/tm/ltm/profile.py
+++ b/f5/bigip/tm/ltm/profile.py
@@ -26,7 +26,6 @@ REST Kind
     ``tm:ltm:profile*``
 """
 
-import logging
 
 from f5.bigip.resource import Collection
 from f5.bigip.resource import OrganizingCollection
@@ -110,8 +109,10 @@ class Client_Ssl(Resource):
             'tm:ltm:profile:client-ssl:client-sslstate'
 
     def create(self, **kwargs):
-        """Allows creation of SSL profile with just key/cert names
-        otherwise it will just require 'name' kwarg to be provided
+        """Allows creation of SSL profile with just key/cert names.
+
+            Will require 'name' kwarg to be provided otherwise
+
         """
         if 'certname' and 'keyname' in kwargs:
             kwargs['name'] = kwargs['certname'][:-4]
@@ -120,7 +121,6 @@ class Client_Ssl(Resource):
 
         self._create(**kwargs)
         return self
-
 
 
 class Analytics_s(Collection):
--- a/f5/bigip/tm/ltm/profile.py
+++ b/f5/bigip/tm/ltm/profile.py
@@ -38,8 +38,60 @@ class Profile(OrganizingCollection):
     def __init__(self, ltm):
         super(Profile, self).__init__(ltm)
         self._meta_data['allowed_lazy_attributes'] = [
-            Client_Ssls
-        ]
+            Analytics_s,
+            Certificate_Authoritys,
+            Classifications,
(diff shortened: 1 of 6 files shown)


zancas:
Hmmm...   some of your tests reference 'web_securitys', 'xmls', 'udps', etc...   but the `Profile` object doesn't have the corresponding attribute.

wojtek0806:
the attributes are there at the very bottom

wojtek0806:
See #349 latest comment

f5/bigip/tm/ltm/profile.py
```diff
@@ -56,9 +108,1070 @@ def __init__(self, client_ssls):
...
+
+            Will require 'name' kwarg to be provided otherwise
+
+        """
+        if 'certname' and 'keyname' in kwargs:
+            certname = kwargs.pop('certname')
+            keyname = kwargs.pop('keyname')
+            kwargs['certKeyChain'] =\
+                [{'cert': certname, 'key': keyname,
+                  'name': certname.split('/')[2][:-4]}]
```
zancas: This needs to be it's own parameter, e.g. `'name': kwargs.pop('name')`.
It should _NOT_ have any constrained relationship with `certname`.

f5/bigip/tm/ltm/profile.py
```diff
@@ -56,9 +108,1070 @@ def __init__(self, client_ssls):
...
+            if 'name' not in kwargs:
+                kwargs['name'] = certname[:-4]
```
zancas: Remove this `if` clause.  i.e. lines 123 and 124.

f5/bigip/tm/ltm/profile.py
```diff
@@ -56,9 +108,1070 @@ def __init__(self, client_ssls):
...
-                   "key": keyname}
-        logging.debug(payload)
-        return self._create(**payload)
+    def create(self, **kwargs):
+        """Allows creation of SSL profile with just key/cert names.
+
+            Will require 'name' kwarg to be provided otherwise
+
+        """
+        if 'certname' and 'keyname' in kwargs:
```
zancas: Let's change the structure here:

`certname`, `keyname`, and `name` are now required arguments.

So let's put something like this in place:

https://gist.github.com/zancas/aab07faf9e55b03e027fc111f1f71d6d

f5/bigip/tm/ltm/profile.py
wojtek0806: why do you even hard code things like certname, keyname? why not just promote proper syntax ? we can have this nice and vanilla like I had initially: where we just used bog standard create method, where only required parameter was 'name:'

This way the user controls the rest

zancas:
Assuming tests pass, etc...

The removal of a default for `name` and the inappropriate relationship between `name` and `certname` are the only remaining blockers for this PR.

Issue #378, opened 2016-05-06
Given the new command mixin, some classes coding need revision


wojtek0806:
As we have introduced command mixin, some of the classes coding needs to be updated to incorporate that.

wojtek0806:
The following contain usage of command: and would benefit from the new command mixin:
- f5\bigip\cm__init__.py
- f5\bigip\sys\config.py
- f5\bigip\sys\failover.py

wojtek0806:
Closed by mistake

wojtek0806:
Issue Review:

The following contain usage of command: and would benefit from the new command mixin:

```
f5\bigip\cm__init__.py
f5\bigip\sys\config.py
f5\bigip\sys\failover.py
```

There are few other existing resources that will benefit from having an option of running a specific command on them i.e reset-stats . This needs some more planning and review against TMSH

wojtek0806:
We can also add to the list:

f5\bigip\sys\crypto

wojtek0806:
Looking at failover.py:

```
    def toggle_standby(self, **kwargs):
        '''Toggle the standby status of a traffic group.

        WARNING: This method which used POST obtains json keys from the device
        that are not available in the response to a GET against the same URI.

        Unique to refresh/GET:
        u"apiRawValues"
        u"selfLink"
        Unique to toggle_standby/POST:
        u"command"
        u"standby"
        u"traffic-group"
        '''
        state = kwargs.pop('state')
        trafficgroup = kwargs.pop('trafficgroup')
        if kwargs:
            raise TypeError('Unexpected **kwargs: %r' % kwargs)
        payload = {u"command": u"run",
                   u"standby": state,
                   u"traffic-group": trafficgroup}
        standby_session = self._meta_data["bigip"]._meta_data["icr_session"]
        uri = self._meta_data['uri']
        response = standby_session.post(uri, json=payload)
        self._local_update(response.json())

```

This should be amended to mimick tmsh:

```
root@(bigip1)(cfg-sync Standalone)(Eval:Active)(/Common)(tmos)# run sys failover ?
Properties:
  "{"      Optional delimiter
  offline  Changes the status of a unit or cluster to Forced Offline. If persist or no-persist are not specified, the change in status will be persisted in-between
           system restarts.
  online   Changes the status of a unit or cluster from Forced Offline to either Active or Standby, depending upon the status of the other unit or cluster in a
           redundant pair.
[...]

PR #432, opened 2016-05-19
Commandmixin.variables

### LINKED ISSUE #381: Include the supported commands as an attribute list?
Sometimes a resource might support more than one command, for example config can either use command:load and command:save, therefore it would be useful to store this kind of information as a metadata element, in a similiar fashion to the ['allowed lazy attributes'] element.

wojtek0806:
Issues:
 Fixes #381 also addreses some of the files listed in #378

f5\bigip\cm__init__.py
f5\bigip\sys\config.py

Problem:
Since we have introduced CommandMixin the allowed commands were not stored as part of meta data.

Analysis:
This patch adds the metadata ._meta_data['allowed_commands'] into ResourceBase class. The reason behind it is that commands can be run against some Collections as well as Resources.
There are few other classes which have hard coded command in the json payload which need to be changed later to fully implement fox for #378 

Tests:
Flake8
Functional Tests


zancas:
This looks pretty good, but it needs tests!

swormke:
@zancas has been super busy, so I looked at this and am going to merge it because looks like you two have worked on it together and it is pretty straight forward.

Issue #437, opened 2016-05-23
Functional test failures in the sys.config tests


pjbreaux:
The functional test to save the config for the sys organizing collection is failing as shown below. The module with the issue is likely: f5/bigip/tm/sys/config

Here's the error:

=========================================================== test session starts ============================================================
platform linux2 -- Python 2.7.6, pytest-2.9.1, py-1.4.31, pluggy-0.3.1
rootdir: /root/common, inifile:
plugins: symbols-0.1.0-a3, cov-2.2.1
collected 1 items

test_config.py F

================================================================= FAILURES =================================================================
___________________________________________________________ TestConfig.test_save ___________________________________________________________

self = <test.functional.tm.sys.test_config.TestConfig object at 0x7fe5f1f23050>, bigip = <f5.bigip.BigIP object at 0x7fe5f1f230d0>

```
def test_save(self, bigip):
    c = bigip.sys.config
```

> ```
>   c.save()
> ```

test_config.py:20:

---

container = <f5.bigip.tm.sys.config.Config object at 0x7fe5f1f23650>, name = 'save'

```
def __getattr__(container, name):
    # ensure this object supports lazy attrs.
    cls_name = container.__class__.__name__
    if 'allowed_lazy_attributes' not in container._meta_data:
        error_message = ('"allowed_lazy_attributes" not in',
                         'container._meta_data for class %s' % cls_name)
        raise LazyAttributesRequired(error_message)

    # ensure the requested attr is present
    lower_attr_names =\
        [la.__name__.lower() for la in
            container._meta_data['allowed_lazy_attributes']]
    if name not in lower_attr_names:
        error_message = "'%s' object has no attribute '%s'"\
            % (container.__class__, name)
```

> ```
>       raise AttributeError(error_message)
> ```
> 
> E           AttributeError: '<class 'f5.bigip.tm.sys.config.Config'>' object has no attribute 'save'

../../../../f5/bigip/mixins.py:95: AttributeError
========================================================= 1 failed in 0.04 seconds =========================================================
root@836b3b0120ae:~/common/test/functional/tm/sys#

zancas:
Hmmm...

caphrim007:
@pjbreaux i noticed the same

pjbreaux:
Sorry that output sucks. Does anyone know how to paste it with no formatting?

zancas:
@pjbreaux here's the potato

wojtek0806:
@zancas 

[...]

PR #441, opened 2016-05-24
Fixes #435, #437

### LINKED ISSUE #435: Allow File and SSL Resource Naming
First off, thanks for creating this! It really helps out.

When creating SSL Certificate/Keys and SSL Profiles, the name of the resource is always set to `certname[:-4]`. See [here](https://github.com/F5Networks/f5-common-python/blob/0.1/f5/bigip/tm/sys/crypto.py#L76-L80), [here](https://github.com/F5Networks/f5-common-python/blob/0.1/f5/bigip/tm/sys/crypto.py#L52-L56), and [here](https://github.com/F5Networks/f5-common-python/blob/0.1/f5/bigip/tm/ltm/profile.py#L59-L62).

I'd like to be able to name of these resources at creation. This is possible from the API and I'd much rather not have to run a modified version just to create the SSL resources with arbitrary names.

wojtek0806:
Problem:  Some of the code contained hardcoded payloads,  and custom methods, given the new command mixin this has been removed causing some tests to fail.
Cert and Key collections did not allow defining certificate names

Analysis: Custom methods from crypto/cert and crypto/key have been removed to be more in line with how the SDK deals with commands.
Config endpoint had load parameter removed due to some unexpected errors. All relevant tests have been corrected to reflect that change

CAVEAT: Cert and key resources need to have certain methods removed/appended to be in line with REST documentation. sys/config endpoint needs to have load command support added at some point. This is to be done in next update

Tests:
Flake8
Functional Tests


wojtek0806:
I am open to questions regarding the above. There was another way we could resolve the crypto key and crypto cert naming problem, without the removal of the custom methods. But given the fact we discussed this yesterday, we will remain in-line with rest of the code. I will add an example docstring to explain one thing. in a min as i see i have missed that

zancas:
@pjbreaux @wojtek0806 has worked hard on this PR, any chance you can get to it today?

PR #449, opened 2016-05-25
Fixes #448

### LINKED ISSUE #448: Refeactor profile tests so they pass on 11.6
Refactor some of tests so they are either skipped or passed in 11.6

wojtek0806:
Added skip options to some profiles as they are available on v12 and up. Some profiles have added properties in v12, those needed to be changes so the tested properties are available on both versions.

This needs to go in before my other PR #441


zancas:
I think we should add version support ASAP, then test appropriately.

pjbreaux:
Reassigning to @zancas. You may need to assist @wojtek0806 in getting the branches straight since development changed.

PR #455, opened 2016-05-26
TMSH run util for general bash, imish, and reboot commands

jgruber:
Added some simple tmsh run util commands we needed for cert clean and issue bogus bash/tmsh commands.

Need until and integration tests.

### DIFF
--- a/f5/bigip/tm/util/__init__.py
+++ b/f5/bigip/tm/util/__init__.py
@@ -1,0 +1,102 @@
+# coding=utf-8
+#
+# Copyright 2015 F5 Networks Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""BIG-IP® Util (util) module
+
+REST URI
+    ``http://localhost/mgmt/tm/util/``
+
+GUI Path
+    ``Util``
+
+REST Kind
+    ``tm:util:*``
+"""
+import json
+
+from f5.bigip.resource import PathElement
+from f5.bigip.mixins import CommandExecutionMixin
+
+
+class UtilExecutionException(Exception):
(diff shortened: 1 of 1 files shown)


swormke:
@jgruber Need to fix the flake8 errors on your branch.

'''
The command "flake8 --exclude=docs ." exited with 1.
'''

swormke:
Hi @jgruber thanks for the contribution.  @zancas is working on getting the developer guide done which will help with some of my comments, but I thought I would get you started.

First I would look at the file `f5/bigip/tm/sys/config.py` because I think this is a very similar example to what you want to do.  Maybe @pjbreaux has a different opinion, but I think this is pretty close.
1. Like you said you must have at least functional tests to get the PR accepted, but we also would like unit tests if they apply.
2. You need to add documentation for the new modules in the docs/ directory.  It is pretty self explanatory if you look in there what you need to do (don't for get to build the docs and view them locally).
3.  You need to add docstrings for the new class and it's public functions
4. This is a general comment for the SDK. Your function calls should probably not be returning anything and should be either updating the object's parameters or something like that.  Lets talk about what you need from the return values and we can decide how to do that, but I don't think it applies if you look at the example file I listed above.
5. You do not need to check the return value of `self._meta_data['icr_session'].post()` because it handles all of that and will raise an `Exception` on its own.  The caller of your `Util.bash()` should be catching exceptions and part of that exception is the response object from the `post()`.  Same as before this probably doesn't apply when using the `CommandExecutionMixin()` appropriately.

pjbreaux:
Just as an update, I'm a +1 on all of the above comments. Also, you are generating a new uri, such as 'bash', which is a new path element under util. May be worth thinking about it being a new class. I'd like to see what @zancas thinks.

pjbreaux:
You can find an example of this, I believe, in https://github.com/F5Networks/f5-common-python/blob/development/f5/bigip/shared/file_transfer.py

wojtek0806:
Is this PR being worked on at the moment? if not I will close it, you can always reopen once you update the code considering the hints and comments above

pjbreaux:
Do not close it.

jasonrahm:
I think this PR can be closed now @pjbreaux, unless there is something in the code or comments you wish to maintain?  I have already begun this work in other PRs against issue #538, some of which has already been merged.

jasonrahm:
is there a need for this PR to remain open @pjbreaux? I'm working my way through the util functions on issue #538 and have what I need from this PR.

wojtek0806:
guys what is happening with this? since @jasonrahm is doing this already, we should close it.
@zancas @pjbreaux

jasonrahm:
agreed it should be closed

wojtek0806:
closing this PR

PR #459, opened 2016-05-27
Add interfaces for client-side string upload

### LINKED ISSUE #454: Add upload buffer method to bigip.shared.file_transfer.uploads
The f5-sdk currently supports file upload to BIG-IP systems (bigip.shared.file_transfer.uploads.upload_file(). We need an interface that supports uploading an in-memory buffer. With the current implementation of f5-sdk we first must write our in-memory buffer to a temporary file and then call upload_file. We need an interface that allows us to pass in a buffer and write the buffer directly to the REST API, preventing a temporary file copy.

zancas:
Issues:
Fixes #454

Problem: We need to upload secrets without writing them to a file
on the client side.

Tests: unittests in test_file_uploads.py

### DIFF
--- a/f5/bigip/shared/file_transfer.py
+++ b/f5/bigip/shared/file_transfer.py
@@ -17,6 +17,7 @@
 #
 
 import os
+from StringIO import StringIO
 
 from f5.bigip.mixins import FileUploadMixin
 from f5.bigip.resource import PathElement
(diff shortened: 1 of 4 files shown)


caphrim007:
Build
82
${BUILD_RESULT}

f5/bigip/shared/file_transfer.py
```diff
@@ -17,6 +17,7 @@
 #
 
 import os
+from StringIO import StringIO
```
pjbreaux, line 20: Could attempt cPython's StringIO, as shown in https://pymotw.com/2/StringIO/#example

In case the files are yuuge.

PR #463, opened 2016-05-31
Versioning change

wojtek0806:
This PR is opened for further work on the issue of versioning change. Mainly designing tests. There were unexpected issues with existing unit tests, we will work on testing further here.

### DIFF
--- a/f5/bigip/__init__.py
+++ b/f5/bigip/__init__.py
@@ -30,7 +30,6 @@ from f5.bigip.tm.shared import Shared as TmShared
 from f5.bigip.tm.sys import Sys
 from f5.bigip.tm import Tm
 from f5.bigip.tm.transaction import Transactions
-from f5.bigip.mixins import ToDictMixin
 
 
 class ManagementRoot(PathElement):
@@ -58,7 +57,7 @@ class ManagementRoot(PathElement):
             'password': password,
             'tmos_version': None,
         }
-        self._ver_set()
+        self._get_tmos_version()
 
     @property
     def hostname(self):
@@ -72,7 +71,7 @@ class ManagementRoot(PathElement):
     def tmos_version(self):
         return self._meta_data['tmos_version']
 
-    def _ver_set(self):
+    def _get_tmos_version(self):
         connect = self._meta_data['bigip']._meta_data['icr_session']
         base_uri = self._meta_data['uri'] + 'tm/sys/'
         response = connect.get(base_uri)
--- a/f5/bigip/__init__.py
+++ b/f5/bigip/__init__.py
@@ -56,6 +56,7 @@ class ManagementRoot(OrganizingCollection):
             'icontrol_version': icontrol_version,
             'tmos_version': None,
         }
+        self._version()
 
     @property
     def hostname(self):
@@ -66,13 +67,12 @@ class ManagementRoot(OrganizingCollection):
         return self._meta_data['icontrol_version']
 
     def _version(self):
-        connect = self._meta_data['bigip']._meta_data['icr_session']
-        base_uri = self._meta_data['uri']
-        endpoint = 'tm/sys/'
-        uri = base_uri + endpoint
-        response = connect.get(uri)
-        ver = response.json()
-        return str(ver['selfLink']).split('=')[1]
+        mgmt_rt = self
+        dvcs = mgmt_rt.tm.cm.devices.get_collection()
+        for i in range(len(dvcs)):
+            if str(dvcs[i].selfDevice) == 'true':
+                tmsh_v = str(dvcs[i].version)
+        self._meta_data['tmos_version'] = tmsh_v
 
 
 class BigIP(ManagementRoot):
--- a/f5/bigip/__init__.py
+++ b/f5/bigip/__init__.py
@@ -18,6 +18,7 @@
 
 
 from icontrol.session import iControlRESTSession
+from urlparse import urlparse, parse_qs
 
 from f5.bigip.cm import Cm
 from f5.bigip.resource import PathElement
@@ -82,7 +83,7 @@ class ManagementRoot(PathElement):
             base_uri = self._meta_data['uri'] + 'tm/sys/'
         response = connect.get(base_uri)
         ver = response.json()
-        version = str(ver['selfLink']).split('=')[1]
+        version = str(parse_qs(urlparse(ver['selfLink']).query)['ver'][0])
         self._meta_data['tmos_version'] = version
 
 
--- a/f5/bigip/__init__.py
+++ b/f5/bigip/__init__.py
@@ -58,7 +58,7 @@ class ManagementRoot(PathElement):
             'password': password,
             'tmos_version': None,
         }
-        self._version()
+        self._ver_set()
 
     @property
     def hostname(self):
@@ -68,13 +68,17 @@ class ManagementRoot(PathElement):
     def icontrol_version(self):
         return self._meta_data['icontrol_version']
 
-    def _version(self):
+    @property
+    def tmos_version(self):
+        return self._meta_data['tmos_version']
+
+    def _ver_set(self):
         mgmt_rt = self
         dvcs = mgmt_rt.tm.cm.devices.get_collection()
         for i in range(len(dvcs)):
             if str(dvcs[i].selfDevice) == 'true':
                 tmsh_v = str(dvcs[i].version)
-        self._meta_data['tmos_version'] = tmsh_v
+                self._meta_data['tmos_version'] = tmsh_v
 
 
 class BigIP(ManagementRoot):
--- a/f5/bigip/__init__.py
+++ b/f5/bigip/__init__.py
@@ -56,7 +56,7 @@ class ManagementRoot(OrganizingCollection):
             'icontrol_version': icontrol_version,
             'tmos_version': None,
         }
-        self._version()
+        self._ver_set()
 
     @property
     def hostname(self):
@@ -66,13 +66,17 @@ class ManagementRoot(OrganizingCollection):
     def icontrol_version(self):
         return self._meta_data['icontrol_version']
 
-    def _version(self):
+    @property
+    def tmos_version(self):
+        return self._meta_data['tmos_version']
+
+    def _ver_set(self):
         mgmt_rt = self
         dvcs = mgmt_rt.tm.cm.devices.get_collection()
         for i in range(len(dvcs)):
             if str(dvcs[i].selfDevice) == 'true':
                 tmsh_v = str(dvcs[i].version)
-        self._meta_data['tmos_version'] = tmsh_v
+                self._meta_data['tmos_version'] = tmsh_v
 
 
 class BigIP(ManagementRoot):
--- a/f5/bigip/__init__.py
+++ b/f5/bigip/__init__.py
@@ -30,6 +30,7 @@ from f5.bigip.tm.shared import Shared as TmShared
 from f5.bigip.tm.sys import Sys
 from f5.bigip.tm import Tm
 from f5.bigip.tm.transaction import Transactions
+from f5.bigip.mixins import ToDictMixin
 
 
 class ManagementRoot(OrganizingCollection):
@@ -53,6 +54,7 @@ class ManagementRoot(OrganizingCollection):
             'local_ip': None,
             'bigip': self,
             'icontrol_version': icontrol_version,
+            'tmos_version': None,
         }
 
     @property
@@ -62,6 +64,15 @@ class ManagementRoot(OrganizingCollection):
     @property
     def icontrol_version(self):
         return self._meta_data['icontrol_version']
+
+    def _version(self):
+        connect = self._meta_data['bigip']._meta_data['icr_session']
+        base_uri = self._meta_data['uri']
+        endpoint = 'tm/sys/'
+        uri = base_uri + endpoint
+        response = connect.get(uri)
+        ver = response.json()
+        return str(ver['selfLink']).split('=')[1]
 
 
 class BigIP(ManagementRoot):
--- a/f5/bigip/__init__.py
+++ b/f5/bigip/__init__.py
@@ -71,12 +71,12 @@ class ManagementRoot(OrganizingCollection):
         return self._meta_data['tmos_version']
 
     def _ver_set(self):
-        mgmt_rt = self
-        dvcs = mgmt_rt.tm.cm.devices.get_collection()
-        for i in range(len(dvcs)):
-            if str(dvcs[i].selfDevice) == 'true':
-                tmsh_v = str(dvcs[i].version)
-                self._meta_data['tmos_version'] = tmsh_v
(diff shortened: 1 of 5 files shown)


wojtek0806:
@zancas: I do not necessarily agree with the below:

> Remove the check for the existence of the tmos_version key. We should GUARANTEE the existence of that key as part of the interface to ManagementRoot. I.E. no "if tmos_v: instantiation of ManagementRoot 

I have implemented _ver_set() method in ManagementRoot as per above(which means we utilize our iControl to obtain the TMOS by parsing selfLink response. (it has been renamed to _get_tmos_version() btw).

However this has caused us problems with some existing unit tests. For example the tests in here:
- f5\bigip\test\test___init__.py

My proposal to this will be to remove the _get_tmos_version() method out of the ManagementRoot and implement it inside the LazyAttributeMixin? 

Or we can leave it as it is and rewrite the unit tests that fail. Depending on the decision here I can start writing some unit tests for this feature as well as amending some functional tests for the trust and profile sub-modules.

wojtek0806:
@zancas @swormke :
we could also do this where we just remove the following from ManagementRoot **init** :

```
self._get_tmos_version()
```

and just call it from _check_supported_versions method inside LazyAttributeMixin

wojtek0806:
Ok I have something different, thanks to Lukasz's help with @property:

```
    @property
    def tmos_version(self):
        # We return tmos_version, or if None,
        # we grab it and return value via _get_tmos_version(),
        # at the same time updating _meta_data['tmos_version']
        return self._meta_data['tmos_version'] or self._get_tmos_version()
```

On my way to test this :)

wojtek0806:
Looks like all of the tests are failing, this is not looking good

wojtek0806:
I have been able to narrow down the problems to 2 root causes: 

Issue 1: Problems with 'tmos_version' property where it is not set on some unit tests. Due to us faking instantiation of ManagementRoot(BIGIP):

Resolution: 

```

    @property
    def tmos_version(self):
        # We return tmos_version, or if None,
        # we grab it and return value via _get_tmos_version(),
        # at the same time updating _meta_data['tmos_version']
        return self._meta_data['tmos_version'] or self._get_tmos_version()

```

In addition to the above you would need to go through the tests we have and add manually the following:

```
instance._meta_data['tmos_version'] = '11.6.0'
```

Here instance will be whatever fake BIGIP or MR object we create.

Issue 2: Sometimes the link to get tmos version gets duplicated so instead of getting:
https://yourhost:443/mgmt/tm/sys/ we get https://yourhost:443/mgmt/tm/sys/tm/sys. 

This is due to the following:

```

class BigIP(ManagementRoot):
    """A shim class used to access the default config resources in 'mgmt/tm.'

    This class is solely implemented for backwards compatibility.
    """
    def __init__(self, hostname, username, password, **kwargs):
        super(BigIP, self).__init__(hostname, username, password, **kwargs)
        self._meta_data['uri'] = self._meta_data['uri'] + 'tm/'
        self._meta_data['allowed_lazy_attributes'] =\
            [TmAuth, TmCm, Ltm, Net, TmShared, Sys, Transactions]
```

Lots of our tests utilize BigIP(host, user, pass) syntax which explains why this is happening, as we append the /mgmt/ with tm/:

```
self._meta_data['uri'] = self._meta_data['uri'] + 'tm/'

```

our MR method: _get_tmos_version does this: 

```
base_uri = self._meta_data['uri'] + 'tm/sys/'
```

I will need to think on how to address, some uri processing to check for existence of tm/ maybe ? As adding a method into BIGIP just for that is not the right thing to do.

wojtek0806:
Doing something like this. should do the trick

```

        if self._meta_data['uri'].endswith('tm/'):
            base_uri = self._meta_data['uri'] + 'sys/'
        else:
            base_uri = self._meta_data['uri'] + 'tm/sys/'

```

f5/bigip/tm/cm/trust.py
```diff
@@ -31,13 +31,17 @@ class Add_To_Trust(UnnamedResourceMixin, ExclusiveAttributesMixin,
 
     def __init__(self, cm):
         super(Add_To_Trust, self).__init__(cm)
+        base_uri = type(self).__name__.replace('_', '-').lower()
```
zancas, line 35: We should never manually manipulate classes-into-uris that should always be done in a single place:

https://github.com/F5Networks/f5-common-python/blob/development/f5/bigip/resource.py#L165

To be clear, we should always and only use this --^ method.

f5/bigip/tm/cm/trust.py
```diff
@@ -54,8 +58,12 @@ class Remove_From_Trust(UnnamedResourceMixin, CommandExecutionMixin, Resource):
 
     def __init__(self, cm):
         super(Remove_From_Trust, self).__init__(cm)
+        base_uri = type(self).__name__.replace('_', '-').lower()
```
zancas, line 62: See above line note.

zancas:
@wojtek0806 do you have some design insight to share?

'''
@zancas: I do not necessarily agree with the below:

```
Remove the check for the existence of the tmos_version key. We should GUARANTEE the existence of that key as part of the interface to ManagementRoot. I.E. no "if tmos_v: instantiation of ManagementRoot
```

'''

swormke:
I agree with @zancas we should guarantee that the key is always there since we have control over this part of the code..  We can do this by simply adding it as `""` or something in the `__init__()`.  It makes writing the code so much cleaner later.

zancas:
One other design question I have for this issue is whether or not there's utility to making the `supported_versions` be a class vs. instance variable of the concrete_subclasses.

As we now understand class attributes are accessible in the `LazyAttributesMixin` since that method has references to the actual classes.

Some reasons we might prefer a class attribute over an instance attribute:
1.   It's a closer mapping to system we're modeling.   Specifically there is 1 class (type) per terminal URI, all instances of that class, should have the same `supported_versions` in other words, I think `supported_versions` is a class attributes, not an instance attribute.
2.   Instantiating and object...   just to see if we should really be instantiating it...   feels hacky.   Or more technically, is risky, particularly if `__init__` has extra-object side effects.    Mostly we've avoided that kind of side-effect (anti-pattern), but we decided to add exactly that kind of side effect in this very feature (and I think it's a good idea!), because we now have a communication between client and device during `ManagementRoot` instantiation!
   
   To be clear, I'm not saying it's clearly the case that we should do this, but now is probably the right time to spend a minute thinking about trade-offs.

zancas:
@wojtek0806 instead of changing `_get_tmos_version` to support tests that use the old `BigIP` pattern, can we change the failing tests to use the new `ManagementRoot` pattern?

The `BigIP` pattern is no longer best-practice, we shouldn't be adding complexity to the rest of the code-base to support it.

f5/bigip/__init__.py
```diff
@@ -65,6 +66,25 @@ def hostname(self):
     def icontrol_version(self):
         return self._meta_data['icontrol_version']
 
+    @property
+    def tmos_version(self):
+        # We return tmos_version, or if None,
+        # we grab it and return value via _get_tmos_version(),
+        # at the same time updating _meta_data['tmos_version']
+        return self._meta_data['tmos_version'] or self._get_tmos_version()
```
zancas, line 75: Nice.

f5/bigip/__init__.py
```diff
@@ -65,6 +66,25 @@ def hostname(self):
...
+
+    def _get_tmos_version(self):
+        connect = self._meta_data['bigip']._meta_data['icr_session']
+        # Need to add this due to BIGIP class URI manipulation
+        if self._meta_data['uri'].endswith('tm/'):
+            base_uri = self._meta_data['uri'] + 'sys/'
+        else:
+            base_uri = self._meta_data['uri'] + 'tm/sys/'
```
zancas, line 83: I prefer to change tests to use `MangementRoot`, instead of adding this conditional just to support `BigIP` <--(deprecated).

f5/bigip/__init__.py
```diff
@@ -65,6 +66,25 @@ def hostname(self):
...
+        response = connect.get(base_uri)
+        ver = response.json()
+        version = str(ver['selfLink']).split('=')[1]
```
zancas: Good, but we should probably use `urlsplit` instead of custom string parsing.

f5/bigip/mixins.py
```diff
@@ -101,10 +109,20 @@ def __getattr__(container, name):
             if name == lazy_attribute.__name__.lower():
                 attribute = lazy_attribute(container)
                 bases = [base.__name__ for base in lazy_attribute.__bases__]
+                # Doing version check per each resource
+                container._check_supported_versions(attribute)
```
zancas, line 113: This looks really sweet to me!   I like it.

f5/bigip/tm/cm/trust.py
wojtek0806, line 62: Regrettably this will not work due to custom endpoint. I had this using get uri from pathelement and it was failing. Try it :)

f5/bigip/__init__.py
wojtek0806: Why would you want to use external lib where pythin builtin is much faster and cleaner to use..it is a string which needs stripping at '='

f5/bigip/__init__.py
wojtek0806, line 83: This is not for tests onlt. This will break backwards compatibility for EVERY use of BigIP..meaning all solutions that use old syntax will get Uri errors..as get tmos version method will fail as it adds tm/sys into Metadata URI

zancas, line 83: Ahhh...   OK then how about we move the special handling into the `BigIP` class?    Probably easiest:

override `_get_tmos_version` in `BigIP`.

f5/bigip/__init__.py
zancas: It's considered best practice to use the community supported tool, rather than reinvent the wheel.   The `urlparse` library is part of the standard lib.   I think the main reason for this is that core standard library logic is extremely well tested/written/understood/maintained.

zancas: https://docs.python.org/2/library/urlparse.html

Note: the exact syntax changes from Python 2 to Python 3.

f5/bigip/tm/cm/trust.py
zancas, line 62: 0.o  ...   why not?   What's the uri?

wojtek0806, line 62: Here is the error and command used to invoke it:

https://gist.github.com/wojtek0806/249f35ac98c4a636779eee5fdb9edec8#file-gistfile1-txt

f5/bigip/__init__.py
wojtek0806: Unless the inheritance mechanisms changed, how is the method  _get_tmos_version in BigIP be inherited by ManagementRoot? If one uses correct syntax i.e mr = ManagementRoot('host', 'user', 'pass) this will break horribly.

One way around it, is to use my previous method where I utilized SDK to obtain device's version. This way we will not have any self link parsing etc, but we will have to bypass tmos version check first time (as you remember my previous method looked like this):
````
    @property
    def version(self):
        mgmt_rt = self
        dvcs = mgmt_rt.tm.cm.devices.get_collection()
        for i in range(len(dvcs)):
            if str(dvcs[i].selfDevice) == 'true':
                ver = str(dvcs[i].version)
                self._meta_data['tmos_version'] = ver
````

Hence my earlier comment:

> @zancas: I do not necessarily agree with the below:

    Remove the check for the existence of the tmos_version key. We should GUARANTEE the existence of that key as part of the interface to ManagementRoot. I.E. no "if tmos_v: instantiation of ManagementRoot

The other way will be to specify a 'uri' method, which is really just breaking up the base_uri logic into 2 methods for each class, for me it will be just moving code around which does not change anything.

File: f5/bigip/__init__.py (hunk 5 above)
Comment (wojtek0806) [inline f5/bigip/__init__.py]: fine but str split works the same in python 3, as per:

https://docs.python.org/3/library/stdtypes.html#str.split

Also string split butilt in method is not what i would call reinventing the wheel.

Comment (wojtek0806) [inline f5/bigip/__init__.py]:
```
version = str(parse_qs(urlparse(ver['selfLink']).query)['ver'][0])
```

f5/bigip/__init__.py
wojtek0806, line 83: I just removed my previous comment as I saw you want to override this in BIGIP, while yea we can do this but why?

why do you want to create another method instead of using the if 2 line statements? What is the point of retyping 90% of the method in another class ?

wojtek0806:
> One other design question I have for this issue is whether or not there's utility to making the supported_versions be a class vs. instance variable of the concrete_subclasses.(...)

There is a major drawback to this, it leads to manually defining and changing parts of the code with each added TMOS version. For example:
- We define "supported_versions = ['11.6.0', '12.0.0'] " class attribute inside PathElement (this is required as some classes do not inherit from ResourceBase, Licensing class in /shared for example)
- Then in the classes that do not support version 11.6.0, we need to define their own class variable:
  "supported_versions = ['12.0.0']". 
- Later on we add more versions to "supported_versions" class attribute inside PathElement, now, we have to go to each class that had their own "supported_versions" defined and edit the list, this has to be done each time we add new version.

With the current solution we just define "self._meta_data['supported_versions'] = ['11.6.0', '12.0.0']" in the PathElement, next we just do the following in classes that do not support 11.6.0:

"self._meta_data['supported_versions'].remove('11.6.0')"

Afterwards adding new classes to supported_versions will not require us to go back and change the above in each class that has been defined.

This is much easier to control as it grows and more scallable. Unless you know of another way we can do this then I am all ears :)

pjbreaux:
If a uri such as Add_To_Trust only exists in 12.0.0, does this change preclude someone who is using 12.0.1 from using that feature, since we are matching the version exactly? And if so, would we have to extend the list for each patch made to the version?

caphrim007:
There's the distutils StrictVersion class that would make than unnecessary 

http://stackoverflow.com/questions/11887762/how-to-compare-version-style-strings

Because we could use >= <= as need to check a range of versions

wojtek0806:
In the previous form you dont need to change anywhere but one place. we have global list where we can add versions as they come. in add to trust we remove those that we don support

wojtek0806:
@pjbreaux Not really, the way we obtain devices version is by extracting that from the Selflink, and from what I have seen in the current tick-tock software releases 11.6.1 and 12.1 that is reflected in the links. However hotfixes are not. I am going to look at the link @tar posted to see how we can do this with current logic of the version checking. I will probably nuke this PR along with the fork and do a clean one anywway. Thanks for input tho.

wojtek0806:
@caphrim007 For doing this comparison you still need to maintain the list of versions, for each class. While this is cool lib, it will not help here. I have been playing with this lib but won't do for what we want to do here.

@zancas  I have been playing about with class level attributes and it feels unmanageable in the long run. I am willing to risk it and have one list to rule them all, and just remove unsupported versions on sub-class level.

wojtek0806:
nuking this PR in favor of the new one.

PR #472, opened 2016-06-01
Feature.active standby cluster

pjbreaux:
@zancas 
#### What's this change do?

Improves documentation and code for supporting an active standby cluster.
#### Any background context?

After talking to jgruber, I realized that having a four-member cluster isn't terribly useful because you still have to manually manage traffic groups. Without this traffic group feature, only an active standby cluster is useful because it uses one traffic group to manage traffic.

Start at the new unit tests and then move to the functional tests. The four-member tests have been removed, since testing against two devices is easier than having four running.

### DIFF
--- a/f5/multi_device/cluster/__init__.py
+++ b/f5/multi_device/cluster/__init__.py
@@ -156,6 +156,7 @@ class ClusterManager(object):
         if hasattr(self, 'cluster'):
             msg = 'The ClusterManager is already managing a cluster.'
             raise AlreadyManagingCluster(msg)
+        self._check_device_number(kwargs['devices'])
         print('Adding trusted peers to root BigIP...')
         self.trust_domain.create(
             devices=kwargs['devices'],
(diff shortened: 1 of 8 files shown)


zancas:
Some checks failing...

pjbreaux:
Travis looks good. Jenkins is not.

caphrim007:
@pjbreaux re-triggering the build here

zancas:
@pjbreaux please sync your development branch with origin and rebase on top of it.

caphrim007:
@pjbreaux why were the scale up/downs removed?

pjbreaux:
A few reasons. Adding a device causes a failover event. Also, as I understand it, having more than two devices usually requires multiple traffic groups meaning we would have to manage those for multiple devices as well as manage their fallbacks. We will add it eventually.

caphrim007:
@smooth-alg see @pjbreaux 's response

pjbreaux:
I'd love to discuss with anyone curious. At least getting requirements around clustering for more than two devices.

Issue #462, opened 2016-05-31
Add version support to ltm profiles


wojtek0806:
Some profiles were not implemented in version 11.6.0. SDK needs to be able to check that

wojtek0806:
implemented in 0.1.7

Issue #461, opened 2016-05-31
Add_To_Trust and Remove_From_Trust classes should never run on 11.6.0 version


wojtek0806:
API needs to return exception if an attempt to instantiate classes is made when the tmos version running on bigip is not 12.x

wojtek0806:
implemented in 0.1.7

Issue #452, opened 2016-05-26
Add_To_Trust and Remove_From_Trust classes URI endpoints broken


wojtek0806:
Need to readd URI endpoints so the classes function properly.

wojtek0806:
temp fix in 0.1.7 need to look at root cause to fix in 0.1.8

wojtek0806:
Root cause was that the classes were inheriting from resource which removes URIs upon **init**. Those classes did not need to inherit from Resource so after debating this with @za we will move this to inherit from PathElement

wojtek0806:
added to 0.1.8 closing

PR #473, opened 2016-06-02
Multiversion Support

### LINKED ISSUE #379: Add version checking to some SDK elements
Some of the functionalities are not available in versions < 12.0, therefore they should not attempt to execute if utilized with BIGIP versions < 12

wojtek0806:
Fixes #379, #452, #461, #462

Problem: SDK needs to be able to run on multiple versions of TMOS. Some API endpoints were not implemented in the base 11.6.0, but exist in 12.x. This SDK needs the ability to verify the TMOS version when instantiating the API classes.

Analysis: This update forces the tmos version update upon ManagementRoot instantiation. Moreover it adds a new private method into LazyAttributeMixin class: _check_supported_versions which is responsible for veryfing if an instantiated class for the API is supported in the target's device tmos.

Files affected:

f5\bigip__init__.py
f5\bigip\mixins.py
f5\bigip\resource.py
f5\bigip\tm\cm\trust.py
f5\bigip\tm\ltm\profile.py

CAVEAT:  This change will affect all of the unit tests that utilize fake instance of BIGIP or ManagementRoot clases. So those tests need rewriting to accommodate the change.

### DIFF
--- a/f5/bigip/__init__.py
+++ b/f5/bigip/__init__.py
@@ -18,6 +18,9 @@
 
 
 from icontrol.session import iControlRESTSession
+from urlparse import parse_qs
+from urlparse import urlparse
+
 
 from f5.bigip.cm import Cm
 from f5.bigip.resource import PathElement
@@ -54,8 +57,10 @@ class ManagementRoot(PathElement):
             'bigip': self,
             'icontrol_version': icontrol_version,
             'username': username,
-            'password': password
+            'password': password,
+            'tmos_version': None,
         }
+        self._get_tmos_version()
 
     @property
     def hostname(self):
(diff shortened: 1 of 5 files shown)


swormke:
@wojtek0806 and @zancas Probably want to fix the title on this one.  I think it used to have a nice mult-version support title or something.

zancas:
Checks failing.

f5/bigip/__init__.py
```diff
@@ -65,6 +70,18 @@ def hostname(self):
...
+    @property
+    def tmos_version(self):
+        return self._meta_data['tmos_version']
+
+    def _get_tmos_version(self):
+        connect = self._meta_data['bigip']._meta_data['icr_session']
+        base_uri = self._meta_data['uri'] + 'tm/sys/'
+        response = connect.get(base_uri)
+        ver = response.json()
+        version = str(parse_qs(urlparse(ver['selfLink']).query)['ver'][0])
```
zancas, line 82: Do we need the `str(` call here?  I doubt it.

wojtek0806, line 82: you will get unicode rather than string then

wojtek0806, line 82: hold on let me check ;]

f5/bigip/tm/cm/trust.py
```diff
@@ -31,13 +31,17 @@ class Add_To_Trust(UnnamedResourceMixin, ExclusiveAttributesMixin,
 
     def __init__(self, cm):
         super(Add_To_Trust, self).__init__(cm)
+        base_uri = type(self).__name__.replace('_', '-').lower()
+        self._meta_data['uri'] =\
+            self._meta_data['container']._meta_data['uri'] + base_uri + '/'
```
zancas, line 36: Let's use `_set_meta_data_uri` here.   If there's some reason that method isn't working then we need to understand it.

wojtek0806, line 36: we can address this later as this is not the only place you have this logic in the code

zancas:
This is landed by PR #478 .

Issue #485, opened 2016-06-07
remove redundant code within CM submodule


wojtek0806:
remove the unnecessary code for:
self._meta_data['uri']

wojtek0806:
Removing the following:
https://github.com/F5Networks/f5-common-python/blob/development/f5/bigip/tm/cm/sync_status.py#L39

Causes sync_status class to suffer the same issue as #452. As this object does not support anything other than load(), we can move its inheritance to PathElement too.

wojtek0806:
I stand corrected, since we also support refresh, we need to inherit from ResourceBase

wojtek0806:
added to 0.1.8 closing

Issue #484, opened 2016-06-07
add_helper function to PathElement


wojtek0806:
This is required to have the following logic also built in for command_mixins:

```
        key_set = set(kwargs.keys())
        required_minus_received =\
            self._meta_data['required_creation_parameters'] - key_set
        if required_minus_received != set():
            error_message = 'Missing required params: %r'\
                % required_minus_received
            raise MissingRequiredCreationParameter(error_message)
```

wojtek0806:
This will also be used here:

https://github.com/F5Networks/f5-common-python/blob/development/f5/bigip/resource.py#L192

wojtek0806:
added to 0.1.8 closing

Issue #481, opened 2016-06-06
Mutually exclusive attributes not enforced with resource creation


wojtek0806:
ExclusiveAttribute mixin does not prevent from sending mutually exclusive parameters when creating resource

wojtek0806:
Need to think how to apply the fix globally so it covers all use cases. Another thing we can insert into path element i guess

wojtek0806:
@zancas @caphrim007

pjbreaux:
I can imagine something like _check_keys for outgoing requests to sanitize keys.

https://github.com/F5Networks/f5-common-python/blob/development/f5/bigip/resource.py#L219

wojtek0806:
added to 0.1.8 closing

Issue #474, opened 2016-06-02
Add custom method to OCSP_Stapling_Params


wojtek0806:
In line with create method for the class, a custom update method is required

wojtek0806:
added to 0.1.8 closing

PR #486, opened 2016-06-07
Multibugfixes_ #378, #452, #474, #481, #484, #485, #491

### LINKED ISSUE #378: Given the new command mixin, some classes coding need revision

wojtek0806:
Fixes #378, #452, #474, #481, #484, #485, #489, #490, #491 

Problem:
Given the latest additions and changes to the core classes some of the sub-modules coding needed change, also during the process few issues were discovered with create method allowing mutually exclusive attributes to be sent to the BIGIP. Also a decision was made to create a helper class to handle required attributes within different parts of code.

Analysis:

Removed the hardcoded payload in failover.py. Kept the legacy method in failover.py for backwards compatibility. Added 2 new helper methods inside PathElement:,
- _check_exclusive_parameters to handle those exclusive attributes during create/exec_cmd.
- _check_required_parameters - a static method to handle checking for required paremeters

Modified CommandExecutionMixin _exec_cmd method to utilize the 2 new helpers from PathElement, modified Resource _create method to do the same. Modified _check_load_parameters method inside PathElement to call _check_required_parameters. In addition to the above I have moved self._meta_data['exclusive_attributes'] declaration into ResourceBase. Added new dictionary entry for: self._meta_data['required_command_parameters'] into PathElement to accomodate splitting this from required_creation_parameters as per discussion with @za. The self._meta_data['required_command_parameters'] is empty by default, this is to allow utilizing classes where 'name' attribute is not present or required, also to allow utilizing commands by the means of  utilCmdArgs= argument.

e.g.

```
exec_cmd('foocommand', utilCmdArgs='foo bar')
```

Files modified:

f5\bigip\resource.py
f5\bigip\mixins.py
f5\bigip\tm\cm\device_group.py
f5\bigip\tm\cm\sync_status.py
f5\bigip\tm\cm\traffic_group.py
f5\bigip\tm\cm\trust.py
f5\bigip\tm\cm\trust_domain.py
f5\bigip\tm\ltm\profile.py
f5\bigip\tm\sys\failover.py
f5-common-python\test\functional\cm\test_sync_status.py
f5-common-python\test\functional\tm\sys\test_failover.py

New files added:

f5\bigip\tm\sys\test\test_failover.py
f5\bigip\tm\ltm\test\test_profile_ocsp.py


wojtek0806:
@za Would be nice if only the errors could be viewed in travis, scrolling through the entire list is painful

zancas:
Looks good.   Did you mean "helper class" or "helper method"?

wojtek0806:
i meant method :)

f5/bigip/resource.py
```diff
@@ -295,6 +305,37 @@ def _handle_requests_params(self, kwargs):
...
+
+    @staticmethod
+    def _check_required_parameters(rqset, **kwargs):
+        """Helper function to do operation on sets
+
+        ::returns bool and variable
+        """
+        key_set = set(kwargs.keys())
+        required_minus_received = rqset - key_set
+        if required_minus_received != set():
```
pjbreaux, line 360: Why not raise the exception here? Allowing the caller to deal with it in whatever way they choose.

f5/bigip/resource.py
```diff
@@ -295,6 +305,37 @@ def _handle_requests_params(self, kwargs):
             requests_params.update({'params': params})
         return requests_params
 
+    def _check_exclusive_parameters(self, **kwargs):
```
pjbreaux, line 332: Since these are changes in resource.py, I'd like to see new unit tests to make sure the functionality is as we suspect. https://github.com/F5Networks/f5-common-python/blob/development/f5/bigip/test/test_resource.py

pjbreaux, line 332: The idea is that we have 100% code coverage for every new piece of code that goes in. If you haven't done with with pytest, you can install pytest-cov and run it like so: `py.test --cov-report term --cov=resource.py test/test_resource.py`

f5/bigip/resource.py
```diff
@@ -295,6 +305,37 @@ def _handle_requests_params(self, kwargs):
...
+        """Check for mutually exclusive attributes in kwargs.
+
+        :raises ExclusiveAttributesPresent
+        """
+        if len(self._meta_data['exclusive_attributes']) > 0:
```
pjbreaux, line 337: Would it suffice to do `if self._meta_data['exclusive_attributes']:`

wojtek0806, line 337: we need to ensure that list is not empty

pjbreaux, line 337: An empty list is falsy, but you should keep it the way you have it. It is more explicit. Mind changed.

f5/bigip/resource.py
wojtek0806, line 332: yea ill whip some more unit tests that test those tomorrow

f5/bigip/resource.py
wojtek0806, line 360: This helper method is used in different parts of code that have different purposes so they raise different exceptions..

pjbreaux, line 360: The method here would raise MissingRequiredParams(msg) and the caller could then reraise that or raise their own exception with a customized message. My hangup here is that the meaning of return values of True and False may not be explicitly understood by a user of this method. The other thing is that in one case a tuple is returned and in the other, a bool is returned. Raising is a clear indication that something went wrong here, then the caller can wrap that in a try except block and decide to do something based on whether it raised or not.

wojtek0806, line 360: this method is not for use outside the core classes..

pjbreaux, line 360: The idea is that this method is merely checking that things are all good, and it finds that or it doesn't. Getting a return value of True, may not mean much to a user, but raising an exception will mean that something's not right. I think it's just a little too specific of a use here and could be generalized a bit to raise. The other thing I'm thinking about is that it really has no information to return back to the caller that can't be done with an exception.

wojtek0806, line 360: We can do the following:

https://gist.github.com/wojtek0806/5084d2d45bc0036dc188aafa3cae44dd

This will allow us to do something like this when using this helper method:

https://gist.github.com/wojtek0806/10d6af27ad8526c44f9135c3afa3d86b

The problem is we will be losing custom exception names like 'MissingRequiredReadParameter'. We can rectify this by doing the following:

https://gist.github.com/wojtek0806/4255aa0e95f1f83c0bf706c685d5246d

We still loose traceback, there are obviously ways we can get that back but they are not recommended and hacky, and it varies quite between v2 and v3 python. I would just go with the first solution, and just have one generic exception name for missing required attributes, as they point to the same thing, does not matter if it is read, create or command attribute

pjbreaux, line 360: That's a tasty solution. I like it.

wojtek0806, line 360: I have done the change and it breaks all of the other tests...i will do something else and just make the method return true or false, and we raise exceptions based on usage.

wojtek0806:
I have done this differently which does not break any tests.  I think this is best solution, as doing reraise will break anything that the users of this SDK might have coded already in terms of exception catching. Also changed helper class name and docstrings should clear out any confusion.

pjbreaux:
@wojtek0806 
Ok now that your gists are back up and I can see you proposal again, I have an update. I like that the helper function raises MissingRequiredParameter. You should be able to get away with not changing any of the existing unit tests by making MissingRequiredLoadParameter and MissingRequiredCreationParameter inherit from MissingRequiredParameter.

`class MissingRequiredParameter(F5SDKError):
    pass
`
`class MissingRequiredCreationParameter(MissingRequiredParameter):
    pass
`
`class MissingRequiredReadParameter(MissingRequiredParameter):
    pass
`

That way, whomever uses the helper method to check required params will get the generic exception from the helper method and raise something more meaningful, since the caller generally has more context. Whether to modify the error message is up to your discretion. I think if you have more information, you can amend the message, but as long as the original message of missing parameter is recorded, it's fine.

wojtek0806:
I have changed it now:

https://gist.github.com/wojtek0806/5c5094639ab8d59af9d2f230d8120657#file-gistfile1-txt

This way you keep everything confined to PathElement:
1. All the required checks are in one place
2. The helper method is hidden from the users
3. The name of the method and its return values are clearer

pjbreaux:
I still vote for raising in the helper method and I think your first inclination in that direction was correct. If I were to put this in @zancas language, it's that we've already established a pattern (his favorite word!) in our repos of raising an exception and having it bubble up to the caller, where the caller must decide what to do. My hangup still with the gist above is that the helper method has found something wrong and it should raise to indicate that all is not right. The pattern would be that a _check_xyz method should check on some expected state and raise if that state is not found.

swormke:
I agree with @pjbreaux that the established pattern is that an exception is raised if there is an error. We long ago made that decision so that we could pass BIG-IP errors back to the caller to be dealt with by them.  We should not deviate from this as this provides a consistent API.

wojtek0806:
This has been already discussed in the slack channel. The above no longer applies. The helper method now returns list, exceptions are raised in the appropriate methods.

zancas:
Why is this build failing?   We need to fix that before we can think about merging...

zancas:
So it looks like `Failover` is not returning the expected value:

=================================== FAILURES ===================================
__________________________ TestFailover.test_exec_cmd __________________________

```
self = <test.functional.tm.sys.test_failover.TestFailover object at 0x7fb7c6aab990>
bigip = <f5.bigip.BigIP object at 0x7fb7c5f753d0>

    def test_exec_cmd(self, bigip):
        f = bigip.sys.failover
        f.exec_cmd('run', offline=True)
        fl = bigip.sys.failover.load()

>       assert 'Failover forced_offline' in fl.apiRawValues['apiAnonymous']
E       assert 'Failover forced_offline' in 'Failover active for 0d 00:02:32\n
```

zancas:
I assume this is not a new test?   Why would it fail now?   Could it be a timing issue?

wojtek0806:
This is a changed test, but this is a problem with timing sometimes the status does not change right away. I have put comments in the code and added some 2 sec delay. It does help when testing against hardware unit, not sure how does this work with whatever you are using.

wojtek0806:
There are other tests that are failing on v12 but they are unrelated to the changes. The first failure is due to the change between 11.6.0 Final and 12.0.0. I have already fixed it however that change was reverted by removed so we are back to failing in v12:

https://gist.github.com/wojtek0806/175851f02716db979bc3389400e8f923#file-test_device_group-py

This has been also failing in HF5 for 11.6.0

wojtek0806:
Next failure is for failover state resource:

https://gist.github.com/wojtek0806/d12b399d8541cc8099611acac9314fe7

Also we are running into generation bug during our vlan test, this unfortunately cannot be skipped:

https://gist.github.com/wojtek0806/d5a024fab2ae056da6c4adcf2dc61cab

We also fail HTTPD tests:

https://gist.github.com/wojtek0806/ebb8167111ac83cd9d2284e36def6bbf

I will see what i can do to fix those that can be fixed.

wojtek0806:
Also this error occurs in v12:

https://gist.github.com/wojtek0806/f62ce855c540cf2737e2fd25c65cecae

wojtek0806:
I have fixed the v12 issues, and Jenkins failure.

zancas:
Hey all ( @wojtek0806 ), really great work on this.  I'm glad to see the use of LooseVersion.

PR #493, opened 2016-06-17
Issues:

### LINKED ISSUE #492: cannot load the sys ntp resource
when trying to load the resource for editing, it raises the following error.

```
Traceback (most recent call last):
  File "ok.py", line 22, in <module>
    mr.tm.sys.ntp.load()
  File "/Users/trupp/src/virtualenv/ansible-2.0/lib/python2.7/site-packages/f5/bigip/mixins.py", line 175, in load
    return self._load(**kwargs)
  File "/Users/trupp/src/virtualenv/ansible-2.0/lib/python2.7/site-packages/f5/bigip/mixins.py", line 187, in _load
    base_uri = self._meta_data['uri']
KeyError: 'uri'
```

caphrim007:
@zancas @wojtek0806 

Fixes #492
#### What's this change do?

The change corrects the order of inheritance
#### Any background context?

Inheriting from the Resource class overrides the inheritance for the
load method in UnnamedResourceMixin

The result is that when using the ntp endpoint and loading the information,
an exception is raised.

### DIFF
--- a/f5/bigip/tm/sys/ntp.py
+++ b/f5/bigip/tm/sys/ntp.py
@@ -32,7 +32,7 @@ from f5.bigip.resource import Collection
 from f5.bigip.resource import Resource
 
 
-class Ntp(UnnamedResourceMixin, Resource):
+class Ntp(Resource, UnnamedResourceMixin):
     """BIG-IP® system NTP unnamed resource
 
         .. note::


zancas:
I'm surprised this fixes the `KeyError` 'uri' issue.

I think the issue is caused by there not being a `_meta_data['uri']` in "pre-created/-loaded" `Resource`s..

See this line of code:

https://github.com/F5Networks/f5-common-python/blob/development/f5/bigip/resource.py#L664

That line should be evaluated regardless of order-of-inheritance...   after all the mixin doesn't have an `__init__` method.

I wonder if `UnnamedResourceMixin` should be redefined as a subclass of `ResourceBase` called `UnnamedResource`?

caphrim007:
if I give it any class that has a uri defined, it seems to work. I just tried with PathElement

PR #496, opened 2016-06-19
Adds GTM iRule api support

### LINKED ISSUE #494: gtm irules not supported
The SDK has support for LTM irules, but not the same for GTM irules. This should be added.

caphrim007:
@wojtek0806 @zancas 

Fixes #494 
#### What's this change do?

This patch adds support and tests for to the SDK for the GTM iRule API
- test/functional/tm/gtm/test_rule.py
- f5/tm/gtm/test/test_rule.py
#### Any background context?

The SDK currently does not support the GTM iRule API.

### DIFF
--- a/f5/bigip/tm/gtm/test/test_rule.py
+++ b/f5/bigip/tm/gtm/test/test_rule.py
@@ -1,0 +1,52 @@
+# Copyright 2015 F5 Networks Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import mock
+import pytest
+
+from f5.bigip import ManagementRoot
+from f5.bigip.resource import MissingRequiredCreationParameter
+from f5.bigip.tm.gtm.rule import Rule
+
+
+@pytest.fixture
+def FakeRule():
+    fake_rule_s = mock.MagicMock()
+    fake_rule = Rule(fake_rule_s)
+    return fake_rule
+
+
+class TestCreate(object):
+    def test_create_two(self, fakeicontrolsession):
+        b = ManagementRoot('192.168.1.1', 'admin', 'admin')
+        r1 = b.tm.gtm.rules.rule
+        r2 = b.tm.gtm.rules.rule
+        assert r1 is not r2
+
+    def test_create_no_args(self, FakeRule):
+        with pytest.raises(MissingRequiredCreationParameter):
+            FakeRule.create()
+
+    def test_create_partition(self, FakeRule):
+        with pytest.raises(MissingRequiredCreationParameter):
+            FakeRule.create(partition='Common')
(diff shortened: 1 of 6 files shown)


wojtek0806:
Looks good to merge, not sure if i got merge rights tho, leaving for @zancas  to merge

wojtek0806:
I have pulled this PR, and while this is unrelated to this PR, it looks like the 4 sec timeouts are causing some false positives when testing against hardware units:

https://github.com/F5Networks/f5-common-python/blob/development/test/functional/tm/sys/test_failover.py#L65

and:

https://github.com/F5Networks/f5-common-python/blob/development/test/functional/tm/sys/test_failover.py#L79

Can you tell me if they fail for you locally too? I added those previously due to Jenkins failures. For me reducing those to 2 sec fixes problem, but I do not want to cause Jenkins fails with that change

caphrim007:
@wojtek0806 was that second comment meant for a different PR?

wojtek0806:
@caphrim007 I meant this for this PR, i just wanna check if failover passes on your end, if not then we can reduce the timeout values and kill 2 birds with 1 stone, rather than raising second PR just to fix the timing in the test

caphrim007:
@wojtek0806 I had no issues running the tests on this feature.

f5/bigip/tm/gtm/test/test_rule.py
```diff
@@ -0,0 +1,52 @@
...
+@pytest.fixture
+def FakeRule():
+    fake_rule_s = mock.MagicMock()
+    fake_rule = Rule(fake_rule_s)
+    return fake_rule
+
+
+class TestCreate(object):
+    def test_create_two(self, fakeicontrolsession):
+        b = BigIP('192.168.1.1', 'admin', 'admin')
```
pjbreaux: Please update to use the new root ManagementRoot

caphrim007: Is there an example? All the existing tests use BigIP

test/functional/tm/gtm/test_rule.py
```diff
@@ -0,0 +1,178 @@
...
+from requests.exceptions import HTTPError
+
+
+RULE = '''when LB_SELECTED {
+   set wipHost [LB::server addr]
+}
+'''
+
+
+def delete_rule(bigip, name, partition):
```
pjbreaux: Could you also do me another small favor and create another fixture alongside the 'bigip', which is called 'management_root' or 'mgmt_root' which has the new path, and use that instead of the old root?

The new fixture would be in here: https://github.com/F5Networks/f5-common-python/blob/development/conftest.py

f5/bigip/tm/gtm/test/test_rule.py
pjbreaux: I'm doing something like that in clustering: https://github.com/F5Networks/f5-common-python/blob/development/f5/multi_device/cluster/test/test_cluster_manager.py

test/functional/tm/gtm/test_rule.py
caphrim007: ok

pjbreaux: Note, you'll also have to update all calls to gtm to be: mgmt_root.tm.gtm...

caphrim007:
@pjbreaux PR updated

pjbreaux:
Well this is all just delectable. If Jenkins passes. I'll merge.

pjbreaux:
Now was this prioritized for a specific milestone? (I should have asked this first)

caphrim007:
@pjbreaux no, it was just something that came up recently when I was doing ansible work

Issue #411, opened 2016-05-12
The Resource `create` and `load` Aliasing Issue


zancas:
`Resource` `create` and `load` create 2 aliases for the same object in the callers namespace, _IF_ they're called on the right hand side of an assignment statement.

This leads to confusing errors like this:

https://github.com/F5Networks/f5-common-python/issues/407

This "design" issue is being created to provide a single location to discuss the problem and come up with fix.

zancas:
I'll add proposed fixes to this thread next time I think about it.

wojtek0806:
Issue Review: Has this been resolved with #414 PR? If so we can close it. Also this seems like duplicate of #415

zancas:
They're similar, but not duplicate.

pjbreaux:
Punting to you @zancas

zancas:
@pjbreaux can you please open a PR from the branch you're using for this to development?  That way we can gather our comments and code into one place.

PR #499, opened 2016-06-21
Fixing issue 411

pjbreaux:
@zancas 
#### What's this change do?

This is just the code changes for fixing the aliasing issue when creating or loading a resource.
#### Any background context?

Issue #411 has more information about this aliasing issue.

Only the code changes are here. One test change was made to fix one failing unit test, but others remain. This will break the build, but you can at least see the code.


f5/bigip/resource.py
```diff
@@ -742,9 +745,11 @@ def _load(self, **kwargs):
         base_uri = self._meta_data['container']._meta_data['uri']
         kwargs.update(requests_params)
         response = refresh_session.get(base_uri, **kwargs)
-        self._local_update(response.json())
-        self._activate_URI(self.selfLink)
-        return self
+        # Make new instance of self
```
zancas: Hang on this code surprises me!   Why isn't `create` just calling `_create`?  That's the standard pattern...   isn't it?    I might have to `git blame` this code.

PR #502, opened 2016-06-21
Add list of design patterns to README.rst

zancas:
Fixes: #501

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -55,6 +55,76 @@ Usage
         pool_b = mgmt.tm.ltm.pools.pool.load(name='mypool', partition='Common')
         pool_b.delete()
 
+Design Patterns
+~~~~~~~~~~~~~~~
+
+I intend the SDK to be easy to use and easy to hack.  These overarching goals
+have a strong influence on my thinking when I am reviewing contributions, this
+means it is in their own interest that I make them as explicit as possible!
+
+The original interface specification was given to me by Shawn Wormke, who I
+believe was influenced by the Jira and Django projects.  At the time I was
+reading Brett Slatkin's 'Effective Python', and I tried to follow its advice
+where possible.
+
+List of Patterns For Contributing Developers
+--------------------------------------------
+
+#. Hack this list to make it more correct/complete
+    For list additions assign @zancas as the PR reviewer.
+#. The call operator ``()`` means: "Try to communicate with the device."
+    This is a strong contract we offer the consumer of the SDK. If an SDK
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -55,6 +55,76 @@ Usage
...
+#. public-nonpublic pairs
+    e.g. 'create' and '_create' XXX add content here.
+#. Handle known issues in the device REST server.
+    The SDK intends to provide
+    a rational interface to consumers that does the right thing.  This means
+    that one case where it does NOT simply do the stupid thing is when it
+    handles a known idiosyncrasy in the device REST server.  For example, some?
+    resources ignore 'disable' and 'enable' configuration options when they are
+    set to 'False', rather than force a consumer to learn about this quirk in
+    the server, then SDK guesses that '"disable": False' means '"enable": True'
```
caphrim007: "the" instead of "then". "the SDK guesses"

README.rst
```diff
@@ -55,6 +55,76 @@ Usage
...
+
+#. public-nonpublic pairs
+    e.g. 'create' and '_create' XXX add content here.
+#. Handle known issues in the device REST server.
+    The SDK intends to provide
+    a rational interface to consumers that does the right thing.  This means
+    that one case where it does NOT simply do the stupid thing is when it
+    handles a known idiosyncrasy in the device REST server.  For example, some?
+    resources ignore 'disable' and 'enable' configuration options when they are
+    set to 'False', rather than force a consumer to learn about this quirk in
```
caphrim007: I'd end the sentence after False and just start a new one at "Rather"

caphrim007:
curious what your thoughts on what tool (if not the SDK) would be used for smoothing the user experience. For example, the verbosity of the OpenStack client libraries has resulted in the shade library https://pypi.python.org/pypi/shade

zancas:
I think that 'smoothing' needs to be decomposed into distinct subtasks...   some examples:
- contextmanager-based transactions for specific configurations
- reasonable defaults for different use-cases
- convenience interpreters that interpret and present relevant configuration information

So "clustering" is an example of code that I think has a lot of higher-level logic cleanly separated from the rest of the SDK.   In the near future I think we need to rename the 'bigip' directory/subpackage in a way that reflects that it's really the "core" of the SDK, and then we can add sister subpackages/directories like clustering...

So, in short I think that the SDK should grow higher level logic...   in cleanly separated namespaces.

zancas:
We as a group need to figure out how we want to rename the `bigip` directory.

Issue #293, opened 2016-03-22
Support stats for resource objects


swormke:
Feature request asked for by @sebastianmaniak Resource objects have a statistics URI associated with them.  For example a get to: `https://host-vm-15/mgmt/tm/ltm/virtual/~Common~test-vserver/stats` will return JSON with the statistics.

We need to decide how to support this, but there are a few options.
1. Add a `get_stats()` function for the base resource object that returns the json from GET.
2. Make it a proper object that would be called like vserver.stats.<mystat>

zancas:
I lean towards a class.

wojtek0806:
Issue Review: I think the above would warrant a method on VS class resource. One of the reasons is the way you have this working in TMSH, to display VS stats you issue the command:

tmsh show ltm virtual exampleserver raw

Of course the "raw" keyword means that the values are not converted into any units. See below output from TMSH help:

```
show ltm virtual blah ?
Options:
  all-properties              Display all properties for the specified items
  default                     Units are determined based on current values
  detail                      Include statistics for all related items
  exa                         All values are displayed in exa-units
  field-fmt                   Show statistics in field format for the specified items
  gig                         All values are displayed in giga-units
  ip-intelligence-categories  IP Intelligence white and black lists categories.
  kil                         All values are displayed in kilo-units
  meg                         All values are displayed in mega-units
  peta                        All values are displayed in peta-units
  port-misuse                 Port misuse policy statistics
  raw                         No conversions are made to any values
  tera                        All values are displayed in tera-units
  yotta                       All values are displayed in yotta-units
  zetta                       All values are displayed in zetta-units
```

Another argument for this being virtual.py method is that we recently implemented command-mixin which allows us to run 'reset-stats' command on the virtual resource.

wojtek0806:
Adding to the above, We could make this a resource method as you can utilize it in pool stats too:

GET to the following /mgmt/tm/ltm/pool/~Common~examplepool/stats

Returns:

"kind": "tm:ltm:pool:poolstats"

Which has its reflection in TMSH:

show ltm pool examplepool raw

And consequently you can run reset-stats command against this resource.

wojtek0806:
This looks like a bigger job, there are more modules than we currently have but even for those that we already have its impressive, i think gradual implementation of those would be the answer, ill start with pool/vs and snat and take it from there.

https://gist.github.com/wojtek0806/8f784a0884cfa2781e942702b55a2f6f

[...]

PR #505, opened 2016-06-22
Add Stats Module - WIP

wojtek0806:
This PR is opened for the benefit of discussing the proposed solution to #293 . It will have tests added after the approach is agreed upon


f5/bigip/mixins.py
```diff
@@ -316,3 +316,79 @@ def get_device_info(self, bigip):
...
+    def _get_stats(self, rdict):
+        """Helper method to convert dictionary into an object
+
+            Each stat element in the returned json
+            contains a single key:value pair dictionary.
+            Hence this method will allow access to each stat value
+            using python 'dot' notation.
+        """
+
+        class DictConvert(object):
```
dangle, line 347: I'd pull this up out of the function. Right now, isinstance checks will fail and consumers might run into strange edge cases.

dangle, line 347: Instead of creating a new class for each call, you might consider creating a class that inherits from dict and overrides __getattr__. Then you'll get things like isinstance(x, dict) and __eq__ for free.

f5/bigip/mixins.py
```diff
@@ -316,3 +316,79 @@ def get_device_info(self, bigip):
...
+        return rdict
+
+    def _update_dict(self, rdict):
+        """Updates dictionary stats key with class object
+
+        We are only interested in the contents 'entries'
+        key of the returned json, rest is irrelevant
+        for us.
+
+        We still utilize _check_keys method to sanitze
```
dangle, line 387: Typo. "sanitze" -> "sanitize"

f5/bigip/mixins.py
```diff
@@ -316,3 +316,79 @@ def get_device_info(self, bigip):
...
+
+                @param :rdict Dictionary
+                """
+                self.__dict__.update(rdict)
+                for k, v in rdict.items():
+                    if isinstance(v, dict):
+                        self.__dict__[k] = DictConvert(v)
+        return DictConvert(rdict)
+
+    def get_stats(self):
```
dangle, line 362: You could make this a "stats" property the way you do with "stats_raw".

f5/bigip/mixins.py
```diff
@@ -316,3 +316,79 @@ def get_device_info(self, bigip):
...
+
+        We are only interested in the contents 'entries'
+        key of the returned json, rest is irrelevant
+        for us.
+
+        We still utilize _check_keys method to sanitze
+        the values before further processing.
+
+        """
+        sanitized = self._check_keys(rdict.pop('entries'))
```
dangle, line 391: Are you sure this is going to have "entries" in it? You might want to do a sanity check here, or at least catch KeyError and throw something that might have meaning to the user.

f5/bigip/resource.py
```diff
@@ -887,3 +887,5 @@ def exists(self, **kwargs):
             else:
                 raise
         return True
+
```
dangle, line 890: Is this whitespace intentional?

wojtek0806:
deleting this as i need to rethink the implementation from scratch

wojtek0806:
reopening this  until ill create new PR

f5/bigip/mixins.py
zancas, line 362: Is it possible to have cyclic references?

zancas, line 362: Is `ToDictMixin` broken?

f5/bigip/mixins.py
```diff
@@ -316,3 +316,79 @@ def get_device_info(self, bigip):
...
+            return {key.replace('.', '_'): self._key_dot_replace(val)
+                    for key, val in rdict.items()}
+
+    def _get_stats(self, rdict):
+        """Helper method to convert dictionary into an object
+
+            Each stat element in the returned json
+            contains a single key:value pair dictionary.
+            Hence this method will allow access to each stat value
+            using python 'dot' notation.
```
zancas, line 344: Why do you want to do this?

wojtek0806, line 344: see my slack discussion with nuno

f5/bigip/mixins.py
wojtek0806, line 362: I have explored that avenue but it does not do what I need it to do. The end goal is to be able to call:

vs.stats.statName.value

and get the actual integer.

zancas:
@wojtek0806 I mostly like this implementation.   SDK architecture is that every path element of a URI needs to be it's own class that inherits from `PathElement` so I think this class should not be a mixin, it should be a class in `resource.py`.

wojtek0806:
closing PR as new one is opened

PR #514, opened 2016-06-30
Fixing issue 411

### LINKED ISSUE #411: The Resource `create` and `load` Aliasing Issue

zancas:
Fixes #411


f5/bigip/tm/net/tunnels.py
```diff
@@ -32,10 +32,10 @@
 from f5.bigip.resource import Resource
 
 
-class Tunnels_s(OrganizingCollection):
+class TunnelS(OrganizingCollection):
```
pjbreaux, line 35: Are you sure you want to do this?

pjbreaux:
The Jenkins tests passed, but Jenkins does not reflect this. @tar. I'm merging because I'm good on this.

pjbreaux:
Hmm, now it says passed. Suspicious...

PR #523, opened 2016-07-04
Feature.dg

jasonrahm:
@zancas @caphrim007 Adding internal/external data-group support. Had to add sys->file support as well for the external data group, though I only added the data-group collection/resource to file, not the other functions. Will address those in separate issues.


zancas:
This looks good to me.

PR #527, opened 2016-07-05
Test Kick

pjbreaux:
Kicking tests on code changes from 0.1.7 to 0.1.9

Do not merge! We will rebase and force push to 0.1 to preserve linear history in 0.1 branch. This is merely to demonstrate tests are successful.


pjbreaux:
Ok @mattgreene, @zancas and @swormke. If these tests prove successful, I will force push to 0.1 with this branch, leaving development alone. This preserves all commits on development with the exception of the alias code and api changes that break backwards compatibility. I have preserved the 0.1.7 tag. The 0.1.8 tag was never added to 0.1 branch, so I will make the 0.1.9 release directly off of 0.1.

Hopefully that makes sense and keeps the history the way we would like it.

pjbreaux:
Hmm got failures on tunnels and applications. Will look into it tonight.

PR #542, opened 2016-07-07
fix fdb bug, and add test

### LINKED ISSUE #532: error using fdb on > 0.1.8
I'm using the development branch of code and getting the following error when trying to access the network fdb
icontrol.exceptions.iControlUnexpectedHTTPError: 400 Unexpected Error: Bad Request for uri: https://10.1.1.2:443/mgmt/tm/net/fdbs/tunnel/~Common~vxlan-flannel-tun?ver=11.5.0
Text: u'{"code":400,"message":"Found unexpected URI tmapi_mapper/net/fdbs/tunnel/~Common~vxlan-flannel-tun.","errorStack":[],"apiError":1}'

Seems OK in previous versions.  But it's possible I'm not using it correctly.  

Here's the code:

```
            payload = {
                "partition": "Common",
                "name": vxlan_tunnel_name
            }
            tun_fdb = mgmt.tm.net.fdbs.tunnels.tunnel.load(**payload)
```

Would be useful to be able to populate the records as well..  Similar to what is documented at: view-source:https://support.f5.com/kb/en-us/products/big-ip_ltm/manuals/product/tmos-implementations-12-0-0/9.html "manually populating L2 location records"

zancas:
Issues:
Fixes #532

Problem: Fdb was misnamed

Analysis: Name and type of Fdb fixed

Tests: test/functional/tm/net/test_fdb.py

### DIFF
--- a/f5/bigip/tm/net/__init__.py
+++ b/f5/bigip/tm/net/__init__.py
@@ -30,7 +30,7 @@ REST Kind
 from f5.bigip.resource import OrganizingCollection
 from f5.bigip.tm.net.arp import Arps
 from f5.bigip.tm.net.dns_resolver import Dns_Resolvers
-from f5.bigip.tm.net.fdb import Fdbs
+from f5.bigip.tm.net.fdb import Fdb
 from f5.bigip.tm.net.interface import Interfaces
 from f5.bigip.tm.net.route import Routes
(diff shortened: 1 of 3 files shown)


pjbreaux:
If Jenkins says so, we gonna ship it.

PR #545, opened 2016-07-08
Fix tunnels_s --> tunnels bug

### LINKED ISSUE #544: tm/net/tunnels is actually an organizing collection so Tunnels_s --> TunnelS

zancas:
@pjbreaux 
Issues:
Fixes #544

Problem:  The `mgmt/tm/net/tunnels` URI corresponds to an OrganizingCollection so make the relevant concrete subclass inherit from that and name it "TunnelS" to distinguish it from the "Tunnels" collection class that corresponds to `/mgmt/tm/net/tunnels/tunnel`

Analysis:

Tests:


pjbreaux:
If Jenkins passes. We'll ship.

PR #548, opened 2016-07-11
Adds coveralls badge to repo

### LINKED ISSUE #547: add coveralls support
Add coveralls support for code coverage tests.

caphrim007:
Issues:
Fixes #547

Problem:
The f5-sdk codebase can run coverage tests, but there is no easy reporting
of these results.

Analysis:
The Coveralls.io service has been setup for the f5-common-python test. This
change to the README is to add the badge for coveralls so that developers
can easily see what the coverage tests currently result in

Tests:
None

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,10 @@
+[run]
+omit =
+    # omit all functional tests
+    tests/*
+    # omit anything in a test directory anywhere
+    */test/*
+    # omit the devtools
+    devtools/*
+    # omit the conftest?
+    conftest.py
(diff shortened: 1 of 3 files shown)


caphrim007:
@pjbreaux @zancas I think this is now ready to go live

pjbreaux:
Do you think there's any value in seeing how much coverage we have of a particular test file? Might be interesting, as shawn said in slack, to see if a test file has been configured to skip half of its tests.

caphrim007:
I originally had the tests in there, but shawn suggested I exclude them. I can add them back if that's what we want

pjbreaux:
Let's exclude them as suggested, then we'll just add them back in if we ever want. I bet we never will.

PR #558, opened 2016-07-18
Fix nat test for test_create_disabled_false

### LINKED ISSUE #552: During 11.5.4 testing, tm/ltm/nat/test_nat.py::test_create_disabled_false fails
Error message is shown below:

icontrol.exceptions.iControlUnexpectedHTTPError: 400 Unexpected Error: Bad Request for uri: https://10.190.3.58:443/mgmt/tm/ltm/nat/
Text: u'{"code":400,"message":"\"disabled\" may not be specified with \"enabled\"","errorStack":[]}'

After further inspection, it was found that the method https://github.com/F5Networks/f5-common-python/blob/81e35af626579f96d13879a4f02506decb421aef/f5/bigip/tm/ltm/nat.py#L96 sees disabled=False and then sets enabled=True, and 11.5.4 does not allow this. The appropriate behavior is that the method should only keep whatever setting should be True. If disabled=False, enabled should be True and that should be submitted.

pjbreaux:
@zancas 

Issues:
Fixes #552

Problem:
When testing against 11.5.4, the JSON being sent to the device had
disabled=False and enabled=True. Only the truthy value should be sent,
thus enabled=True. Even if the user sets disabled=False, we should
modify the resource to ensure the truthy value is set properly and
remove the other Falsy value.

Analysis:
The _endis_able method was previously defined in the nat module and
datacenter module. This has now been moved into a mixin and is avaiable
to any class.

Tests:
The nat tests pass with 11.5.4 and 11.6.0. I will let Jenkins check the
gtm datacenter tests. I also added a few new tests for test_mixins

### DIFF
--- a/f5/bigip/tm/gtm/datacenter.py
+++ b/f5/bigip/tm/gtm/datacenter.py
@@ -27,6 +27,7 @@ REST Kind
     ``tm:gtm:datacenter:*``
 """
 
+from f5.bigip.mixins import EnDisableMixin
 from f5.bigip.mixins import ExclusiveAttributesMixin
 from f5.bigip.resource import Collection
 from f5.bigip.resource import Resource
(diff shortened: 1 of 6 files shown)


zancas:
Looks good, let's generalize it.

Make a new key in:

 `Resource._meta_data['reduction_forcing_pairs'] = [('vlansEnabled', 'vlansDisabled')]`

Now inside `Resource._create`, `Resource._update`, after the config dict is built (i.e. after kwargs, and self.**dict** are combined), run `_reduce_boolean_pair` for each pair in `self._meta_data['reduction_forcing_pairs']:

like this:

``` Python
for p1, p2 in self._meta_data['reduction_forcing_pairs]:
    conf_dict = self._reduce_boolean_pair(conf_dict, p1, p2)
```

Then in the constructor to `Datacenter` add:

``` Python
self._meta_data['reduction_forcing_pairs'].append(('enabled', disabled))
```

Then rerun test.

f5/bigip/tm/gtm/datacenter.py
```diff
@@ -41,21 +42,16 @@ def __init__(self, gtm):
...
-class Datacenter(Resource, ExclusiveAttributesMixin):
+class Datacenter(
+        Resource, ExclusiveAttributesMixin, ReduceBooleanPairToTrueMixin
+):
     """BIG-IP® GTM datacenter resource"""
     def __init__(self, dc_s):
         super(Datacenter, self).__init__(dc_s)
         self._meta_data['required_json_kind'] =\
             'tm:gtm:datacenter:datacenterstate'
         self._meta_data['exclusive_attributes'].append(('enabled', 'disabled'))
```
zancas, line 51: Perhaps just replace 'exclusive_attributes' with 'reduction_forcing_pairs'.

pjbreaux:
Why vlansEnabled or vlansDisabled? Every resource will then have that as a reduction pair, but for some reason we may not want that. No, the resource itself should add that pair.

PR #563, opened 2016-07-18
Bugfix.user create 11 5 4

pjbreaux:
@zancas 
#### What's this change do?

Fixes a test issue that is only apparent when testing against 11.5.4. The create user test was expecting a field, partitionAccess, to be a dict, but it is a string in 11.5.4.
#### Any background context?

We are fixing up failures that occur when testing against an 11.5.4 device. When you run pytest, you must include the --release option, so the appropriate test will be skipped based on the version given.

Only a test fix.

### DIFF
--- a/test/functional/auth/test_user.py
+++ b/test/functional/auth/test_user.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 #
 
+from distutils.version import LooseVersion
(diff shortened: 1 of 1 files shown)


caphrim007:
tested on 11.5.4 hf1. works

PR #566, opened 2016-07-19
Bugfix.tg 11 5 4

pjbreaux:
@zancas 
#### What's this change do?

Fixes a test issue that is only apparent when testing traffic_group for a specific attribute ('mac') in 11.5.4.
#### Any background context?

The test/functional/cm/test_traffic_group.py::TestTrafficGroups::test_device test was failing against 11.5.4 because the above attribute does not exist for a traffic_group.

Only test fixups here.


zancas:
LGTM!

PR #568, opened 2016-07-19
Fixing functional test issues in test/functional/tm/ltm/test_policy.py

pjbreaux:
@zancas 
Don't merge until #558 has been merged!
#### What's this change do?

Implemented a module-level pytest fixture which snapshots the device
before and after the test. The diff between those snapshots is used to
remove objects created by the tests.
#### Any background context?

During test teardown, an attempt is made to delete the policies that are
not a part of the built-ins. The built-ins have changed between versions
of TMOS. We should snapshot the get_collection call and use that
pre-test snapshot to delete all those policies created by the tests.

conftest.py change, then the changes in the test_policy.py file

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -15,6 +15,7 @@
 
 from f5.bigip import BigIP
 from f5.bigip import ManagementRoot
+from f5.utils.testutils.registrytools import register_device
 import mock
 import pytest
 import requests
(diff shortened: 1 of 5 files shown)


caphrim007:
which bigip versions were showing the failure? all of them?

pjbreaux:
11.5.4

caphrim007:
checking

conftest.py
```diff
@@ -153,3 +154,22 @@ def deleter():
...
+    that persist after suite runs.
+    '''
+
+    before_snapshot = register_device(mgmt_root)
+
+    def teardown():
+        after_snapshot = register_device(mgmt_root)
+        diff = set(after_snapshot) - set(before_snapshot)
+        for item in diff:
+            after_snapshot[item].delete()
```
pjbreaux, line 173: Add try except around this statement.

pjbreaux:
Sorry @tar: reassigning back to @zancas, as he was curious about these changes.

f5/bigip/resource.py
```diff
@@ -568,6 +568,30 @@ def _produce_instance(self, response):
         new_instance._activate_URI(new_instance.selfLink)
         return new_instance
 
+    def _reduce_boolean_pair(self, config_dict, key1, key2):
```
pjbreaux, line 571: @zancas: Moved this method up to ResourceBase because the class Failover is an UnnamedResource, which inherits from ResourceBase, not resource.

PR #569, opened 2016-07-19
Adds Stats Module

wojtek0806:
@zancas @pjbreaux  @caphrim007 

What's this change do?

This PR will add the STATS capability to selected resources. In this first iteration only Virtual Server resources will be supported, more resources will be added gradually. 

Any background context?
BIGIP allows users to retrieve stats on many resource objects in the bigip, this is rough equivalent to tmsh show.... command.

At the moment only Virtual Servers are enabled to support stats, rest of the resources will be added in gradually.
Stats module for collections needs different approach hence this was not implemented in this release, but is planned for the future

f5-common-python\test\functional\tm\ltm\test_virtual.py
f5-common-python\f5\bigip\resource.py
f5-common-python\f5\bigip\test\test_resource.py
f5-common-python\f5\bigip\tm\ltm\virtual.py

### DIFF
--- a/f5/bigip/resource.py
+++ b/f5/bigip/resource.py
@@ -161,6 +161,11 @@ class UnsupportedOperation(F5SDKError):
     pass
 
 
+class InvalidStatsJsonReturned(KeyError):
+    """Returned stats JSON should always contain 'entries' key"""
+    pass
+
+
 class PathElement(LazyAttributeMixin):
     """Base class to represent a URI path element that does not contain data.
 
@@ -187,6 +192,8 @@ class PathElement(LazyAttributeMixin):
         self._meta_data['required_command_parameters'] = set()
         # You can't have more than one of the attributes in any of these sets.
         self._meta_data['exclusive_attributes'] = []
+        # We assume that resource does not have stats, to be overridden in subclasses
+        self._meta_data['object_has_stats'] = False
 
     def _set_meta_data_uri(self):
         base_uri = self.__class__.__name__.lower()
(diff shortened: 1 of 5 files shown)


caphrim007:
@wojtek0806 @zancas and now I'm interested in this module because it appears I need to make use of it (or something like it) right now.

So what needs to be done to resurrect this?

f5/bigip/resource.py
```diff
@@ -161,6 +161,11 @@ class UnsupportedOperation(F5SDKError):
     pass
 
 
+class InvalidStatsJsonReturned(KeyError):
+    """Returned stats JSON should always contain 'entries' key"""
+    pass
+
```
caphrim007, line 167: I dont think this is always the case. Here's an example of exemplary engineering

```
{
  "kind": "tm:sys:service:servicecollectionstats",
  "selfLink": "https://localhost/mgmt/tm/sys/service/stats?ver=11.6.0",
  "apiRawValues": {
    "apiAnonymous": "acctd        down, Not provisioned\naced         down, Not provisioned\nadmcloud     down, Not provisioned\nadmd         down, Not provisioned\nalertd..."
  }
}
```

caphrim007:
and @pjbreaux if he has any thoughts

f5/bigip/resource.py
wojtek0806, line 167: We can handle those kind of exceptions when we start enabling this on each submodule, for now I can see LTM stats follow the pattern I set above.

zancas:
Before I look at this more closely I think I'll fetch the feature branch and rebase it on top of `development`.

f5/bigip/resource.py
```diff
@@ -369,6 +377,48 @@ def raw(self):
         """
         return self.__dict__
 
+    def _key_dot_replace(self, rdict):
```
zancas, line 380: Why is this necessary?   Can we add a test that fails without the presence of this helper?

f5/bigip/resource.py
```diff
@@ -369,6 +377,48 @@ def raw(self):
...
+                tmp_dict = nested_dict['entries']
+                return self._key_dot_replace(tmp_dict)
+
+        return self._key_dot_replace(rdict)
+
+    def _get_stats_raw(self):
+        """Displays JSON object transformed into python dictionary"""
+        read_session = self._meta_data['bigip']._meta_data['icr_session']
+        base_uri = self._meta_data['container']._meta_data['uri'] + 'stats/'
+        response = read_session.get(base_uri)
```
zancas, line 407: Can we not hook into `refresh` here?

f5/bigip/resource.py
```diff
@@ -932,3 +985,42 @@ def load(self, **kwargs):
...
+        super(Stats, self).__init__(container)
+        self._update_stats(self._get_stats_raw())
+
+    @property
+    def stats_raw(self):
+        """Provides JSON object converted to a python dictionary"""
+        return self._get_stats_raw()
+
+
+class DottedDict(dict):
```
zancas, line 1011: Sweet.

f5/bigip/resource.py
```diff
@@ -932,3 +985,42 @@ def load(self, **kwargs):
...
+
+        It also allows the caller to retrieve the converted json
+        response in unprocessed state.
+    """
+    def __init__(self, container):
+        super(Stats, self).__init__(container)
+        self._update_stats(self._get_stats_raw())
+
+    @property
+    def stats_raw(self):
```
zancas, line 1006: I think this is the thing that should be called `raw`.

f5/bigip/resource.py
wojtek0806, line 380: This is required as having dotted keys will prevent you from using this dictionary with dot notation

f5/bigip/resource.py
wojtek0806, line 1006: this will be misleading, raw is a call to **dict** of the class, stats_raw is a json blob that is used for later processing, this was added as an option so the user of the SDK will be able just to grab the json and process it as they see fit

f5/bigip/resource.py
wojtek0806, line 407: I was looking at it but stats work a bit differently, apart from the need to append URI with 'stats/' refresh grabs the json and then performs local_update: 

https://github.com/zancas/f5-common-python/blob/statsmodule/f5/bigip/resource.py#L558

We do not process stats json the same way, if you look at the update_stats helper method in PathElement you probably know what i mean: 

https://github.com/zancas/f5-common-python/blob/statsmodule/f5/bigip/resource.py#L416

Given the fact as @caphrim007 noticed that we might run into situation where the json structure is different, we would need to add/tweak that _update_stats method to fit the situation and cover the inconsistent output (great programming there i must say ;)).

That is why i think we need to evolve this once we start adding more and more resources.

f5/bigip/resource.py
zancas, line 380: As I mention in the rebased PR...  I have reservations about the DottedDict...  why are we adding that?

zancas:
Superseded by #627 .

PR #570, opened 2016-07-19
Bugfix.failover tests 11 5 4

pjbreaux:
@zancas 
Close #558 before this one!
#### What's this change do?

Updates failover resource to use new ReduceBooleanPairToTrueMixin to reduce online and offline in that resource and remove local logic that was doing the same thing as the mixin. Also fixed the tests for this resource to teardown properly, leaving the device in an online state.
#### Any background context?

During functional testing in 11.5.4 device goes to forced offline and does not come back online. We also needed to remove sleeps from these tests to make them more event driven.

Tests, then resource changes.


pjbreaux:
@caphrim007: I've updated this PR with you as the assignee. Let me know if you have the bandwidth to look at it. All changes are for 11.5.4 compatibility.

pjbreaux:
I will address your slack comments. Thanks!

caphrim007:
after those are addressed, it should be ready to merge, but i will test again

Issue #555, opened 2016-07-15
During functional testing in 11.5.4 in ltm/profile some tests fail at setup


pjbreaux:
We have 8 test failures against 11.5.4 and all but one have the same signature:

Found unexpected json pair at module

This often occurs in the HelperTest method setup_test, which does a proactive teardown to ensure the box is clean, yet when querying a nonextant profile on 11.5.4 (profile.exists), the device kicks back a 400.

```
>>> cl = bigip.tm.ltm.profile.client_ldaps.client_ldap
>>> cl.exists(name='nonextant_profile')
/usr/local/lib/python2.7/dist-packages/requests/packages/urllib3/util/ssl_.py:122: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
/usr/local/lib/python2.7/dist-packages/requests/packages/urllib3/connectionpool.py:821: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/root/f5-common-python-1/f5/bigip/resource.py", line 894, in exists
    session.get(base_uri, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/icontrol/session.py", line 230, in wrapper
    raise iControlUnexpectedHTTPError(error_message, response=response)
icontrol.exceptions.iControlUnexpectedHTTPError: 400 Unexpected Error: Bad Request for uri: https://10.190.3.58:443/mgmt/tm/ltm/profile/client-ldap/nonextant_profile
Text: u'{"code":400,"message":"Found unexpected json pair at module /ltm/profile. The json pair is \\"client-ldap\\":{\\"nonextant_profile\\":null}.","errorStack":[]}'
```

pjbreaux:
After further inspection, it's simply that these collections do not exist on the 11.5.4 device. Added minimum tmos version to these collections, so they will not be instantiated by a user.

PR #571, opened 2016-07-20
During functional testing in 11.5.4 in ltm/profile some tests fail at setup

pjbreaux:
@zancas 
#### What's this change do?

After further inspection, it's simply that these collections (such as Client_Ldaps) do not exist on the 11.5.4 device. Added minimum tmos version to these collections, so they will not be instantiated by a user. Added some tests to ensure that the correct exception is raised if a user does try to instantiate one of these collections.
#### Any background context?

We have 8 test failures against 11.5.4 and all but one have the same signature: Found unexpected json pair at module

Start at the test changes and go to the profile module from there.


pjbreaux:
@tar: jenkins won't wake up to tell me what's wrong with him. I have a suspicion, so I'm kicking the build again.

pjbreaux:
@za: If this build succeeds. Don't close this pull request until you close #555 and #570

PR #581, opened 2016-07-26
During functional testing against 12.1.0, sys/application test fails and errors out

pjbreaux:
@caphrim007 
#### What's this change do?

Added return to the super call in the overridden version of _create in
Service. Added checks in each of the functional tests to ensure that the
thing that is returned from a create is not None.
#### Any background context?

This is a bug in application.py where we expected the call to create an
application service to raise an HTTPError, which was happening in
versions prior to 12.1.0. This is now fixed on the BIG-IP and when
testing against 12.1.0, it revealed a bug in our code where (in the
happy path) we weren't returning the result from the call to
super()._create(). That was returning None as the object back from a
create call.

Review the code change, then checkout the functional test changes.

### DIFF
--- a/test/functional/tm/sys/test_sys_application.py
+++ b/test/functional/tm/sys/test_sys_application.py
@@ -13,16 +13,14 @@
 # limitations under the License.
 #
 
-from pprint import pprint as pp
 from requests import HTTPError
 
 from f5.bigip.tm.sys.application import Service
 from f5.bigip.tm.sys.application import Template
 
-pp(__file__)
 TESTDESCRIPTION = 'TESTDESCRIPTION'
 
-# Application Service Dictionary
+# Application service dictionary, which matches JSON structure
 sections = {
     'implementation': '',
     'presentation': ''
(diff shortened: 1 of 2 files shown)


zancas:
Do we have a test that validates that the returned object is not the object that had return called on it?

pjbreaux:
We can and we will!

pjbreaux:
@zancas or @caphrim007: Ok, this has the change suggested above.

test/functional/tm/sys/test_sys_application.py
```diff
@@ -53,19 +56,22 @@ def setup_template_test(request, bigip, name, partition):
...
     def teardown():
         delete_resource(test_template)
+
     request.addfinalizer(teardown)
     test_template = template_s.template.create(
         name=name,
         partition=partition,
         actions=definition
     )
+    assert isinstance(test_template, Template)
```
zancas, line 64: I'd like to see something like:

`assert id(test_template) != id(template_s.template)`

PR #593, opened 2016-08-02
During functional testing against 11.5.4, test/functional/tm/net/test_vlan.py tests fail

### LINKED ISSUE #590: During functional testing against 11.5.4, test/functional/tm/net/test_vlan.py tests fail
The following three tests fail with errors related to tagMode:

`
iControlUnexpectedHTTPError: 400 Unexpected Error: Bad Request for uri: https://10.190.3.38:443/mgmt/tm/net/vlan/~Common~somevlan/interfaces/1.1/
E           Text: u'{"code":400,"message":"Found unexpected json pair at configuration item /net/vlan/~Common~somevlan/interfaces/1.1. The json pair is \\"tagMode\\":\\"service\\".","errorStack":[]}'
`

This should be possible to remedy by overriding create and update in vlan to check for tagMode and raise an exception if found.

pjbreaux:
@zancas 

Issues:
Fixes #590

Problem:
The tagMode argument or attribute is not allowed on TMOS version 11.5.4
and less. We need to fail in a meaningful way when using that version.

Analysis:
Added a check to create and update in vlan to ensure we raise an
exception if tagMode is present in kwargs or as a self attribute.

Tests:
Tests against 11.6.0 and 11.5.4 pass

### DIFF
--- a/f5/bigip/tm/net/vlan.py
+++ b/f5/bigip/tm/net/vlan.py
@@ -31,6 +31,13 @@ from f5.bigip.mixins import ExclusiveAttributesMixin
 from f5.bigip.resource import Collection
 from f5.bigip.resource import MissingUpdateParameter
 from f5.bigip.resource import Resource
+from f5.sdk_exception import F5SDKError
+
+from distutils.version import LooseVersion
+
+
+class TagModeDisallowedForTMOSVersion(F5SDKError):
+    pass
 
 
 class Vlans(Collection):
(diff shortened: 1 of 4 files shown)


caphrim007:
Things like this are probably going to creep up in other modules. Do we need a general purpose test that any resource can use?

pjbreaux:
@caphrim007: I agree, but I think the abstraction needed to call a generic resource with generic kwargs might be difficult to read. I'm okay with having a specific test for each resource. That way it's very explicit what we're testing and why.

pjbreaux:
@caphrim007: I made zancas assignee by sheer habit. Let me know if you have time to look at this.

caphrim007:
Roger that

zancas:
@pjbreaux "Some checks were not successful"

zancas:
Let's mark the failing test "skipif < 11.6.0", and then add another test (later) that tests <11.6.0.

Issue #564, opened 2016-07-18
ssl profiles and parent profile inheritance


waffie1:
I've found what I believe to be a limitation to the current ssl profile implementation.  When I run update on a profile, it sets the custom flag to enable on all attributes, breaking the inheritance from the parent profile on options I did not intend to modify.

Example, take an existing ssl profile, which has the following configuration:

```
ltm profile client-ssl /Common/test {
    app-service none
    cert /Common/mycert.crt
    cert-key-chain {
        mycert_ca-bundle {
            cert /Common/mycert.crt
            chain /Common/ca-bundle.crt
            key /Common/mycert.key
        }
    }
    chain /Common/ca-bundle.crt
    defaults-from /Common/clientssl
    inherit-certkeychain false
    key /Common/mycert.key
    passphrase none
```

Now perform an update via the SDK

```
profile = mr.tm.ltm.profile.client_ssls.client_ssl.load(name='test', partition='Common')
profile.secureRenegotiation = 'require-strict'
profile.update()
```

The result will be that all options are 'checked' in the WebUI, and the resulting config now looks like this:

```
ltm profile client-ssl /Common/test {
    alert-timeout indefinite
    allow-dynamic-record-sizing disabled
    allow-expired-crl disabled
    allow-non-ssl disabled
    app-service none
    authenticate once
    authenticate-depth 9
    ca-file none
    cache-size 262144
    cache-timeout 3600
    .... shortened, you get the idea ...
```

I understand the REST interface doesn't have the inherited flag like SOAP, but you can just PUT the option you want to modify leaving the other options as is.  the update method apparently does a PUT on every option(having no reference to what was inherited and what was not), and thus removes the inherited flag.  This will really hamper the ability to use the update method and preserve parent profile inheritance.

Is there some other way to do this without breaking the inheritance on options I do not want to modify and need to continue to inherit from the parent profile?

Thanks

caphrim007:
hmm. this is a head-scratcher. I need to ask the REST developers about this one as I dont think it's an sdk limitation.

caphrim007:
@waffie1 what version of BIG-IP are you using??

wojtek0806:
@waffie1 Upon reading the REST user guide, it can be concluded that PATCH should be used for such modifications. I will open an issue for that to have this method implemented:

From the REST guide:

https://devcentral.i131.f5.com/d/icontrol-rest-api-user-guide-version-1210

'''
[...]

PR #597, opened 2016-08-02
Change update from put to patch

### LINKED ISSUE #578: SDK should support PATCH method
Lack of such support leads to issues like this : #564

Implementing a version of update method that uses PATCH would be a good approach.

zancas:
@caphrim007 
Please reassign to @pjbreaux , and @wojtek0806 , after review comments...   I want to make sure all core developers look at this before merge.
Issues:
Fixes #578 #564

Problem: The use of PUT in update was causing unintended side-effects.
This is costly behavior..  we probably shouldn't force the user to
submit more data than they intend!

Analysis: I've changed "put" to "patch" in the update method.
This seems like a major change, so I'd appreciate lots of feedback.

Among other things I'd like to validate that the f5-openstack-agent
consumer does _not_ break against this new version.

Tests:
- f5/bigip/test/test_resource.py
- f5/bigip/tm/sys/test/test_sys_application.py

### DIFF
--- a/f5/bigip/resource.py
+++ b/f5/bigip/resource.py
@@ -474,7 +474,7 @@ class ResourceBase(PathElement, ToDictMixin):
 
         data_dict.update(kwargs)
 
-        response = session.put(update_uri, json=data_dict, **requests_params)
+        response = session.patch(update_uri, json=data_dict, **requests_params)
         self._meta_data = temp_meta
         self._local_update(response.json())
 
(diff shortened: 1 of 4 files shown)


zancas:
Reviewers please consider this:

http://restful-api-design.readthedocs.io/en/latest/methods.html

(If you're not already familiar.)

zancas:
https://tools.ietf.org/html/rfc5789

caphrim007:
i think people's intention is probably more patchy that puty. Curious to see if all the tests will pass.

zancas:
@caphrim007 all tests have indeed passed.

zancas:
@pjbreaux I'd like your input on this.

zancas:
I'd like @wojtek0806 's input, but he's away, so I'm OK merging without it.  I've talked with @jlongstaf about this issue and he thinks put --> patch is probably a good idea.

zancas:
This PR is now "on top" of the bugfix.alias PR so that one should be closed first.

zancas:
@caphrim007 I looked at this build failure...  it looks like all of the tests pass...

caphrim007:
@zancas going to rekick it to ensure that I have my infra changes correct

zancas:
OK @pjbreaux I've dropped this into your bailiwick.

PR #600, opened 2016-08-03
add array markup to .travis.yml for file deploy to github

zancas:
@richbrowne


richbrowne:
This is corrected based on experience with the f5-icontrol-rest package.

PR #601, opened 2016-08-03
Update pin

zancas:
@pjbreaux


pjbreaux:
Ran cluster tests again, passed.

PR #604, opened 2016-08-04
Fixes transactions and adds functional tests

### LINKED ISSUE #603: fix broken transactions
At some point, transaction code became broken. I think it happened around the time that the .create() method started returning an object instead of modifying itself in place.

Since there was no functional test for transaction creation, this error slipped through the cracks.

caphrim007:
@zancas @pjbreaux 

Issues:
Fixes #603

Problem:
The transaction classes became non-functional when the create() methods
switched to returning an object instead of modifying the object in place.
Due to a lack of functional testing for transactions, this caused them
to break without any knowledge to us.

Analysis:
This change fixes the transaction classes to handle returned objects instead
of objects that are modified in place. Additionally, it creates a functional
test to prevent this issue from creeping up in the future.

Tests:
tests/functional/tm/transaction/test_transaction.py

### DIFF
--- a/f5/bigip/contexts.py
+++ b/f5/bigip/contexts.py
@@ -47,7 +47,8 @@ class TransactionContextManager(object):
         be able to use the API object as you normally would.
         """
 
-        self.transaction.create()
+        self.transaction = self.transaction.create()
+
         self.icr.session.headers.update({
             'X-F5-REST-Coordination-Id': self.transaction.transId
         })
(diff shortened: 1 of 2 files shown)


pjbreaux:
@caphrim007: what is the criticality of this issue?

caphrim007:
@pjbreaux any code that wants to use transactions will not work without this fix. This includes the ansible modules

caphrim007:
and @jasonrahm and his ssl certificate work

zancas:
Nice catch @caphrim007 !

PR #612, opened 2016-08-09
Feature.modify 609

pjbreaux:
@zancas 
#### What's this change do?

Implements the modify method, which is a requests.patch call to exclusively update that which a user specifies by sending in the keyword arguments. The arguments are those that are updated, and nothing more.

Start at the unit and functional tests, then move on to resource.py changes.

### DIFF
--- a/f5/bigip/resource.py
+++ b/f5/bigip/resource.py
@@ -438,31 +438,31 @@ class ResourceBase(PathElement, ToDictMixin):
         """
         super(ResourceBase, self).__init__(container)
 
-    def _modify(self, **kwargs):
+    def _modify(self, **patch):
         """Wrapped with modify, override in a subclass to customize."""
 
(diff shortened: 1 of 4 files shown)


pjbreaux:
@zancas: check out the failure in jenkins for the 11.6.0 run. This test doesn't fail for me locally, when testing against 12.0.0

zancas:
Is there still a 11.6.0 Jenkins fail?   This PR shows "All checks have passed".

zancas:
@pjbreaux I ran the tests locally, 11.6.0:

`442 passed, 37 skipped in 109.24 seconds`

The skips are mostly `gtm`.

I have some unittests on my version of the feature branch:

`feature.unittest_modify_609`

So, can you either rebase on top of it, or otherwise get it to be the thing this PR is to include it?

pjbreaux:
Rebased and pushed.

PR #615, opened 2016-08-10
Adds snmp endpoints

### LINKED ISSUE #613: snmp endpoints
Endpoints for the snmp functionality should be added including /snmp, /snmp/users, /snmp/communities, and /snmp/traps

caphrim007:
Issues:
Fixes #613

Problem:
The api is missing functionality for manipulating the snmp endpoints

Analysis:
This change adds functionality for the snmp endpoints
- snmp
- snmp/communities
- snmp/traps
- snmp/users

Tests:
functional and unit

### DIFF
--- a/test/functional/tm/sys/test_snmp.py
+++ b/test/functional/tm/sys/test_snmp.py
@@ -1,0 +1,614 @@
+# Copyright 2016 F5 Networks Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+
+import copy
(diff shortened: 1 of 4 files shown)


wojtek0806:
looks good, however I would also add modify test to SNMP endpoint, as it supposed to support PATCH method, unless it does not and docs are incorrect.

test/functional/tm/sys/test_snmp.py
```diff
@@ -0,0 +1,614 @@
...
+            request, mgmt_root, 'Common', 'modtest1'
+        )
+        original_dict = copy.copy(comm1.__dict__)
+        commName = 'communityName'
+        comm1.modify(communityName='footest1')
+        for k, v in original_dict.items():
+            if k != commName:
+                original_dict[k] = comm1.__dict__[k]
+            elif k == commName:
+                comm1.__dict__[k] == 'footest1'
```
wojtek0806: should this not be something like:

```
assert comm1.__dict__[k] == 'footest1' 

```

zancas:
@caphrim007 Is there anything blocking this from merging (other than review)?

test/functional/tm/sys/test_snmp.py
zancas: I think we should add `assert` here like @wojtek0806 suggests.

zancas: @caphrim007 so I think all that remains to be done is to add "`assert`".

zancas:
`assert LGTM`

PR #619, opened 2016-08-10
Add modify tests to all profiles.

### LINKED ISSUE #564: ssl profiles and parent profile inheritance

zancas:
@pjbreaux

NOTE:  Check #564 and verify for yourself, manually, that this version of the code does indeed fix the issue.

After you close this PR, check the verify that #564 is also closed.

Issues:
Fixes #564

Problem: There was no verification that the modify feature fixed issue #564

Analysis: These test enhancements verify that the specific issue is resolved.

Tests: test/functional/tm/ltm/test_profile.py

### DIFF
--- a/f5/bigip/resource.py
+++ b/f5/bigip/resource.py
@@ -14,6 +14,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 #
+
 """This module provides classes that specify how RESTful resources are handled.
 
 THE MOST IMPORTANT THING TO KNOW ABOUT THIS API IS THAT YOU CAN DIRECTLY INFER
--- a/f5/bigip/resource.py
+++ b/f5/bigip/resource.py
@@ -212,18 +212,6 @@ class PathElement(LazyAttributeMixin):
             self._meta_data['container']._meta_data['uri'] + endpoint + '/'
         self._meta_data['uri'] = final_uri
 
-    def _check_load_parameters(self, **kwargs):
-        """Params given to load should at least satisfy required params.
-
-        :params: kwargs
-        :raises: MissingRequiredReadParameter
-        """
-        rset = self._meta_data['required_load_parameters']
-        check = self._missing_required_parameters(rset, **kwargs)
-        if check:
-            error_message = 'Missing required params: %s' % check
-            raise MissingRequiredReadParameter(error_message)
-
     def _check_create_parameters(self, **kwargs):
         """Params given to create should satisfy required params.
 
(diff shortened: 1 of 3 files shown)


pjbreaux:
No failures for me on monitor

pjbreaux:
Can you kick this one more time. I want to see it succeed.

pjbreaux:
Looks like 11.5.4 failure there

zancas:
@pjbreaux I wonder why the `update` test on `Ipother` didn't fail in the same way in `11.5.4`?

zancas:
@pjbreaux I'm punting.  I've wasted too much time on this.

zancas:
@pjbreaux I'm +1 on merging this, now.

pjbreaux:
I'm fine with skipping on 11.5.4, but shouldn't the user be notified (via an exception) if they attempt this in 11.5.4 for Ipother?

zancas:
@pjbreaux yes, see:   https://github.com/F5Networks/f5-common-python/issues/622

PR #624, opened 2016-08-11
Statsmodule2

zancas:
@wojtek0806 this is my proposed changeset for the stats feature.   Feel free to pull it rebase it, etc.   Then we can manage it through your original PR and close this one.  Or we can work off this one, whichever you prefer.

### DIFF
--- a/f5/bigip/resource.py
+++ b/f5/bigip/resource.py
@@ -338,7 +338,7 @@ class PathElement(LazyAttributeMixin):
             temp_dict[key.replace('.', '_')] = value
         return temp_dict
 
-    def _pop_nest_stats(self, rdict):
+    def _get_nest_stats(self, rdict):
         """Helper method to deal with nestedStats
            as json format changed in v12.x
         """
@@ -346,7 +346,8 @@ class PathElement(LazyAttributeMixin):
             check = urlparse.urlparse(x)
             if check.scheme:
                 nested_dict = rdict[x]['nestedStats']
-                return self._key_dot_replace(nested_dict.pop('entries'))
+                tmp_dict = nested_dict['entries']
+                return self._key_dot_replace(tmp_dict)
 
         return self._key_dot_replace(rdict)
 
@@ -364,7 +365,7 @@ class PathElement(LazyAttributeMixin):
             error = 'Missing "entries" key in returned JSON'
             raise InvalidStatsJsonReturned(error)
         sanitized = self._check_keys(rdict)
-        stat_vals = self._pop_nest_stats(sanitized.pop('entries'))
+        stat_vals = self._get_nest_stats(sanitized['entries'])
         temp_meta = self._meta_data
         self.__dict__['stat'] = DottedDict(stat_vals)
         self._meta_data = temp_meta
--- a/f5/bigip/resource.py
+++ b/f5/bigip/resource.py
@@ -229,7 +229,8 @@ class PathElement(LazyAttributeMixin):
         self._meta_data['required_command_parameters'] = set()
         # You can't have more than one of the attributes in any of these sets.
         self._meta_data['exclusive_attributes'] = []
-        # We assume that resource does not have stats, to be overridden in subclasses
+        # We assume that resource does not have stats,
+        # to be overridden in subclasses
         self._meta_data['object_has_stats'] = False
 
     def _set_meta_data_uri(self):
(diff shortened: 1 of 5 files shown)


zancas:
Mostly I just moved things about a bit.

One significant change I made was to make "Stats" be added to the "allowed_lazy_attributes" of _ALL_ Resource instances by default...    I'm not sure how bad of an idea this is.

If a Resource doesn't have stats, then SDK will incorrectly assume it does, try to get it and fail...  probably with a 404.   This violates our policy of preventing known failed operations from ever reaching the wire...   BUT...  it means all Resources that _do_ support stats should automagically work...  "out of the box".

Also I'm not sure about this DottedDict business.   It's definitely cute...   but doesn't it have the potential to confuse the user?   Usually the '.' operator in the SDK implies access of a new URI element?

I think I prefer to force users to access elements of the "stats" object as a dictionary with the '[]' operator.   I think it helps clarify that the 'stats' attribute is a different kind of thing than a resource.

Overall, however, I'm extremely pleased with this feature.

Please don't close it until I've added a few more functional tests.

Please let me know what you think of this version.

zancas:
Hmmm......    there's a class of operations in this code that I think we might want to move out of `resource.py`.

I think `class Stats(PathElement):` should just be responsible for HTTP METHOD wrapping, URI construction management etc.

I think manipulation of the return data structures (which I'm starting to believe can vary significantly) should be handled by entirely different objects.

f5/bigip/resource.py
```diff
@@ -1035,3 +1043,84 @@ def load(self, **kwargs):
...
+        """Displays JSON object transformed into python dictionary"""
+        read_session = self._meta_data['bigip']._meta_data['icr_session']
+        base_uri = self._meta_data['container']._meta_data['uri'] + 'stats/'
+        response = read_session.get(base_uri)
+        rdict = response.json()
+        return rdict
+
+    def _update_stats(self, rdict):
+        """Attaches stat attribute to stats object"""
+        if 'entries' not in rdict:
```
zancas, line 1101: Maybe this kind of logic needs to move into a new class in a different module.

Do we need a `response_handlers.py`?

zancas:
@caphrim007 @pjbreaux I'm working off of @wojtek0806 's `Stats` implementation.

I'm thinking we should be separating concerns here.

I don't yet have a good feel for what stats JSON blobs returned from the device look like.

Mostly the code in `resource.py` doesn't concern itself with the contents of the JSON configurations it receives.

That seems like the domain of other objects....    thoughts?

One thing I've done is turn "Stats" "on" for ALL Resources.

@wojtek0806: I think we should move all post-processing of the JSON stats out of `resource.py`

Tentatively we need a `response_handlers.py` module....

Is that a bad idea?

Where would such a thing live?

@wojtek0806:   I think you should pull my version of your code.

Your code, and commits are still there.

It's rebased on top of the most recent `development`

The tests are refactored a bit.

The helper functions are moved into `Stats` (out of `PathElement`)

I made it the defaults that _ALL_ `Resource` objects assume they have a `Stats` lazy attribute.

Other than those things I haven't changed the original version.

I think we should consider breaking all (or almost all) JSON post-processing out of the `class Stats(PathElement):` class....     and probably put it into a different module entirely.

zancas:
@wojtek0806 I propose a 3 step process:

(1) remove all logic that post-processes JSON
(2) add functional tests of 2 or 3 distinct stats supporting resources that validate that we're getting the stats JSON blob we expect
(3) add logic and a test that shows something TOTALLY SANE happens when we try to access a stats resource that's not there.
(4) merge the above code
(5) open a new issue for post-processing stats JSON
(6) create a new module called `response_handers.py`?
(7) put all stats JSON handling logic in `response_handlers.py`
(8) test the crap out of it
(9) open a subsequent PR for landing the JSON handling logic

zancas:
And by "3" I meant "9"...  ;)

zancas:
@caphrim007 if you could grab "raw" json with e.g.:

`mgmt_rt.FOO.BLAH.SOME_RESOURCE.stats`

would that be useful to you?

caphrim007:
or a .json() method. requests has such a thing

zancas:
`mgmt_rt.FOO.BLAH.SOME_RESOURCE.create('a', 'b').stats.json`?

zancas:
Superseded by #627 .

PR #627, opened 2016-08-12
Feature.stats 293

### LINKED ISSUE #293: Support stats for resource objects

zancas:
@pjbreaux @wojtek0806 

Fixes: #293 
This PR is intended to close #293.   It is "Minimum Viable".   There is extensive useful utlility (created by @wojtek0806 that will extend stats in an "Enhanced Stats Utils" PR).

### DIFF
--- a/f5/bigip/__init__.py
+++ b/f5/bigip/__init__.py
@@ -71,10 +71,6 @@ class ManagementRoot(PathElement):
     def icontrol_version(self):
         return self._meta_data['icontrol_version']
 
-    @property
-    def tmos_version(self):
-        return self._meta_data['tmos_version']
-
     def _get_tmos_version(self):
         connect = self._meta_data['bigip']._meta_data['icr_session']
         base_uri = self._meta_data['uri'] + 'tm/sys/'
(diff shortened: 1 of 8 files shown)


zancas:
Please see #628 for tracking stats tranformations utils.

f5/bigip/resource.py
```diff
@@ -807,6 +771,7 @@ def _activate_URI(self, selfLinkuri):
         # attrs local alias
         attribute_reg = self._meta_data.get('attribute_registry', {})
         attrs = attribute_reg.values()
+        attrs.append(Stats)
```
pjbreaux, line 774: Does this imply that you can have stats on a collection?

f5/bigip/__init__.py
```diff
@@ -71,10 +71,6 @@ def hostname(self):
     def icontrol_version(self):
         return self._meta_data['icontrol_version']
 
-    @property
```
pjbreaux, line 74: Backwards compat was broken with this change. I should have recognized it and we should have released 2.0.0.

PR #639, opened 2016-08-17
Fix broken backwards compat in regards to tmos_version

### LINKED ISSUE #638: Fix broken backwards compat in regards to tmos_version
The tmos_version attribute was changed from being a property to being a key in the _meta_data dictionary on the bigip container. This change broken backwards compat without use realizing immediately, and we need to support both methods until we're ready to bump the release to 2.0.0.

pjbreaux:
@jlongstaf  and @mattgreene 

Issues:
Fixes #638

Problem:
The tmos_version attribute was changed from being a property to being a
key in the _meta_data dictionary on the bigip container. This change
broken backwards compat without use realizing immediately, and we need
to support both methods until we're ready to bump the release to 2.0.0.

Analysis:
Added the @property back to the f5/bigip/**init**.py for tmos_version.
Both ways of accessing tmos_version are now supported

Tests:
Added a test to ensure we can access the tmos_version both ways

### DIFF
--- a/f5/bigip/__init__.py
+++ b/f5/bigip/__init__.py
@@ -71,6 +71,10 @@ class ManagementRoot(PathElement):
     def icontrol_version(self):
         return self._meta_data['icontrol_version']
 
+    @property
+    def tmos_version(self):
+        return self._meta_data['tmos_version']
+
     def _get_tmos_version(self):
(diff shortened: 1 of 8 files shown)


mattgreene:
Ok to commit pending your last investigative test.

pjbreaux:
Merging. Thanks.

PR #661, opened 2016-08-23
Add toggles for modules to symbols file

### LINKED ISSUE #660: support toggling module checks on and off in symbols
The symbols file can be used for toggling module checks on and off for your platform. If, for example, you do not have GTM licensed, then you may not want to run those tests when you run the standard py.test command.

caphrim007:
Issues:
Fixes #660

Problem:
There are cases when you want to run the full test suite, but omit certain
checks based on the lack of modules in your bigip install. There is currently
no way to ignore these tests without getting creative in your test file location
to the py.test command.

Analysis:
This patch adds the ability to specify modules to turn "on" or "off" in a symbols
file when running the funtional tests.

Tests:
none needed

### DIFF
--- a/symbols.example.yaml
+++ b/symbols.example.yaml
@@ -1,0 +1,9 @@
+---
+
+bigip_netloc: "localhost"
+bigip_port: 10445
+bigip_username: "admin"
+bigip_password: "admin"
+modules:
+    gtm: false
+    ltm: true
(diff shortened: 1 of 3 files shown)


pjbreaux:
Are those the only places where we expect certain modules to be provisioned?

caphrim007:
those are the gtm provisions

pjbreaux:
I'd like to either create an issue to go find the other tests that require a certain module, if any, and add the same logic to those, or fix them in this pull request and send out a note to others to tell them to now mark any module that is needed in the symbols file.

Issue #634, opened 2016-08-17
add virtual server support to GTM endpoint


wojtek0806:
lack of this part blocks GTM  pool sub-collections testing. Need to add this before GTM pools are fully functional

Issue #633, opened 2016-08-17
add wideip support to GTM endpoint


wojtek0806:
this is required in order to implement gtm pools correctly

PR #664, opened 2016-08-24
Adds GTM wideip endpoints

caphrim007:
Issues:
Fixes #663

Problem:
GTM Wide IP endpoints were not available.

Analysis:
Adds the endpoints to support wideips on both 11.x and 12.x

Tests:
functional and unit

PR #670, opened 2016-08-27
Adding GTM pools, wideips and servers

### LINKED ISSUE #621: gtm pools
GTM pools api does not exist in the sdk. It would be nice to have this so that ansible code could be ported to use it.

wojtek0806:
Fixes #621, #633, #634

Problem:
The pool, wideip and server GTM endpoints should be added in the gtm organizing collection.
They nature of the pool and wideip collections have changed significantly between v11.x and v12.x.
In v12.x both pool and wideip have become Organizing collections.
The result was breaking changes occurred in the API.
Servers endpoint was required as a dependency for pools

Analysis:
In order to maintain consistent user experience we have overloaded **new**
method inside the class factory for both pool and wideip endpoints.
Result now being that the SDK automatically detects which object to return
based on the BIG-IP you are communicating with.
Server endpoint was added as it was a dependency for GTM pools

Tests:
Flake8
Functional Tests
Unit Tests

Files Added/Changed:

f5-common-python/f5/bigip/tm/gtm/**init**.py
f5-common-python/f5/bigip/tm/gtm/pool.py
f5-common-python/f5/bigip/tm/gtm/wideip.py
f5-common-python/f5/bigip/tm/gtm/server.py

f5-common-python/f5/bigip/tm/gtm/test/test_pool.py
f5-common-python/f5/bigip/tm/gtm/test/test_wideip.py
f5-common-python/f5/bigip/tm/gtm/test/test.server.py

f5-common-python/test/functional/tm/gtm/test_pool.py
f5-common-python/test/functional/tm/gtm/test_wideip.py
f5-common-python/test/functional/tm/gtm/test_server.py


wojtek0806:
I tagged all 3 of you depending which one of you has bandwidth to review
@zancas @pjbreaux @caphrim007

caphrim007:
Tested and works on the following platforms. Haven't had a chance to test others.
- 11.6.0
- 11.6.1

My problem with this PR is the Members and the Members_s classes of the GTM pools code. It seems to me to break all of the existing conventions we have for Resources, Collections, and OrganizingCollections. I don't think it should be in the business of being a big conditional statement. I think those resources should be individual classes.

It's current design makes me think "premature optimization", whether that was its intention or not. That design though, limits our flexibility with modifying these resources in the future, and makes future maintenance difficult. If any refactoring needs to take place or overriding functionality (consider _update or _create, etc) needs to happen for a specific endpoint to address breaking functionality in the underlying API, this design limits us.

I also think the introduction of a new meta_data variable (v12_gtm_pool_type) is a bad idea. It's used in this case solely for the purposes of addressing a _deficiency_ in the classes design. That suggests bad design and therefore a need to correct it.

Meta data really should be concerned with more general purpose stuff and less class specific stuff. What happens if it changes in v12.1 or v13. Do we add more metadata to cover those cases? I think appropriately named classes can easily be created out of these conditional statements, and then to handle their "different" class names (to provide a consistent user experience) one can simply override the class name as shown here

https://github.com/F5Networks/f5-common-python/pull/664/files#diff-d78a02f174da6910dd417302d09c5aa9R78

So I would look for those changes (unless I'm convinced otherwise) before I would +1 the PR.

Soliciting @zancas and @pjbreaux opinions as well.

zancas:
I agree with @caphrim007 's analysis concerning conventional naming.   Following conventions/using established patterns makes our code easier to read...   and readability is extremely important, if we want this thing to stay alive, so we should adhere to established patterns wherever possible!!

Also, I agree with @pjbreaux ,  it's _possible_ to break this work into multiple smaller features, each of which is easier for a reviewer to digest...     making a reviewers life easier should always be in the back of our minds...   after all, we are the reviewers!   So let's make our own lives easier with smaller PRs.   Rule of thumb:  Once you've changed 5 modules for a feature (including test, etc.) ask yourself, can I package up the work I've done into it's own PR?!?!?

If yes, then do!

A smaller PR is _ALWAYS_ better (except when it's not).

zancas:
@jasonrahm Please read the preceding comment, I'd like for you to be aware also!

jasonrahm:
Thanks @zancas, reading these threads several times over. Good for my personal growth to observer these conversations.

wojtek0806:
@zancas @pjbreaux We cannot break this PR down to smaller ones, those features depend eachother during testing, therefore it will remain as it is, there is a list of files that have been added in the PR notes and only in those files you should be looking at the code. As far as naming conventions go, well we are not going by convention here, as we are overloading **new** method,  I will play with members_s,  but all you do is move code around, those if statements will be there in one form or another, tho they might be easier to read.

wojtek0806:
added code for review, i would need some help tho in figuring out how to do unit tests on sub-collections so we have better coverage

zancas:
Huh...   this still seems like 3 features to me.

wojtek0806:
Pools depend on WideIPs and Servers, WideIPs depend on DataCenters and Servers, you cannot have 3 PRs that will have working tests without all of the endpoints present. Just wait till we start implementing APM, then you will see more than 3 features in PR that simply need to be there

jasonrahm:
The GTM object dependency hierarchy is wideip->pool->vips->server->datacenter, so if you reverse that list with ordered pull requests, and just add the functional tests iteratively as PRs are merged, it shouldn't be a problem, though certainly extra work decoupling existing functional tests.

caphrim007:
My issue with that argument is that with the exception of datacenters (because we already have them in the SDK) you can have them as 3 different PR's because you enter them in to the BIG-IP as 3 different steps.

Where I think we went astray in this PR is we thought that multiple features in a PR was a good way to do business; when it is not. Consider any other open source project and the maintainers will no doubt reject PRs that encapsulate multiple features. That's what they were asking for here.

While the argument that "well if they all go in at the same time so that's the same as them each going in individually" is probably technically correct, it also makes the discussion about any of the code more scattered. We need to track discussion on 3 or more different topics instead of focusing on just one.

That's my view of the problem that the maintainers are trying to address here by asking for separate PRs. I wouldn't be able to get away with it in any other project that I contribute to, and one wouldn't get away with it when doing any BIG-IP development. So in the interests of maintaining focus and providing critical commentary about individual features, that's why it should be in separate commits.

Note that separate commits does not preclude the ability to merge everything at the same time. git is more than happy to branch off of a local branch you have that uses a previous feature you worked on as it's parent. During the commit process all of those can go in. But during the review process it's better if they are all separate.

As I'm not familiar with APM, is there a specific example endpoint I could reference on my local box?

wojtek0806:
i will break it apart but without functional tests for pools and possibly wideips. We add those tests in 4th PR. I don't have time to take stuff apart any other way

PR #689, opened 2016-09-13
Fix unordered dict bug in to_dict()

### LINKED ISSUE #687: iterating in to_dict() causes test failures on python3
The `to_dict` method, a mixin, defined at `ToDictMixin` has a bug where intermittently it will fail on python3.

caphrim007:
Issues:
Fixes #687

Problem:
On python3 (and also I presume on python2) the test_two_refs test in
test_mixins.py will fail because the returned value is

```
{"x": [TraversalRecord, "z"], "z": [1, "a"]}
```

instead of the expected value

```
{"x": [1, "a"], "z": ["TraversalRecord", "x"]}
```

Analysis:
I tracked this down to the `_traverse_dict()` method where it iterates
over the values of the `instance_dict` variable. The problem is that
because dictionaries are unordered, the first value that is iterated
upon will sometimes return "z" first instead of "x".

This causes the result to become "turned around" and the test to fail.

I never saw this crop up on python 2, but it crops up regularly on
python 3. The fix I applied is to cast the unordered dict in `_traverse_dict`
into an OrderedDict. Doing this makes the values of that variable always
known, so-to-speak, because they have a defined order. Therefore the
tests will pass as expected.

Tests:
- f5/bigip/test/test_mixins.py::test_to_refs

### DIFF
--- a/f5/bigip/mixins.py
+++ b/f5/bigip/mixins.py
@@ -17,7 +17,14 @@
 # NOTE:  Code taken from Effective Python Item 26
 
 
+try:
+    from collections import OrderedDict
+except ImportError:
+    from ordereddict import OrderedDict
+
 from distutils.version import LooseVersion
+from six import iteritems
+
 import logging
 
 from f5.sdk_exception import F5SDKError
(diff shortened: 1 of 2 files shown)


caphrim007:
@zancas and @pjbreaux this requires some architectural discussion as it adds a new pip dependency in python 2.6.

zancas:
Nice job @caphrim007 !!   Thanks.  I'm +1 to merge this...     I'm OK with a new pip dependency for <= 2.6.

Issue #538, opened 2016-07-06
Add tm/util subpackage


jasonrahm:
The cli diagnostic and system commands available to tmsh like:
- unix-mv
- unix-rm
- ping
- traceroute
- dig

and others are missing. The unix-rm especially is necessary for cleaning up files uploaded while importing iFiles/External Data-Groups via iControl REST.

caphrim007:
@jasonrahm any idea if any work has gone into this?

jasonrahm:
In my quest to learn git and clean up some mistakes, I blew away some work I had stashed. But I was planning on doing some work with this one after I land the datagroup tests.

Jason

caphrim007:
k

jasonrahm:
Looks like @jgruber's #455 PR has some of the code I was working on for this issue as well. It is tagged help wanted, so I'll pull his PR to my local box and start working on this issue later today.

jasonrahm:
For discussion...please see https://github.com/F5Networks/f5-common-python/pull/671/ for details. Basically, these commands all return 200 whether successful to end-user intent or not. The commandResult field needs to be evaluated for success/failure. Is that something we should take on or leave to the end-user to solve? Note that the data in commandResult is just raw text. Thoughts appreciated.

jasonrahm:
speaking of conventions @zancas, looking for direction on this front. Some of this conversation is in in general between @caphrim007 and me. Just not sure how I should handle providing errors to the users of the sdk since none of these commands return anything but a 200 response. The error information if present is in the commandResult field. Thanks for any advice, haven't found any similar issues pecking through the current state of the sdk.

wojtek0806:
merged so closing

PR #690, opened 2016-09-13
Issue #538 - Adding bash utility

jasonrahm:
Updated file:
f5/bigip/tm/util/**init**.py

Added files:
f5/bigip/tm/util/Bash.py
f5/bigip/tm/util/test/test_bash.py
test/functional/tm/util/test_bash.py

### DIFF
--- a/f5/bigip/tm/util/Bash.py
+++ b/f5/bigip/tm/util/Bash.py
@@ -1,0 +1,65 @@
+# coding=utf-8
+#
+# Copyright 2016 F5 Networks Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+"""BIG-IP® utility module
+
+REST URI
+    ``http://localhost/mgmt/tm/util/bash``
+
+GUI Path
+    N/A
+
+REST Kind
+    ``tm:util:bash:*``
+"""
+
+from f5.bigip.mixins import CommandExecutionMixin
+from f5.bigip.resource import UnnamedResource
+from f5.utils.util_exceptions import UtilError
+
+
+class Bash(UnnamedResource, CommandExecutionMixin):
+    """BIG-IP® utility command
+
+    .. note::
+
+        This is an unnamed resource so it has not ~Partition~Name pattern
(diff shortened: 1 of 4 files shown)


jasonrahm:
@caphrim007 and @zancas this is ready for review. 

Structurally I believe it is sound, but I wonder if we should allow any string in utilCmdArgs that doesn't start with "-c ..." since any other use of bash from REST would be useless. Thoughts appreciated.

jasonrahm:
@pjbreaux & @zancas, since you're in the neighborhood today, and thoughts on my previous comment?

pjbreaux:
What would happen if I sent in FakeiControl.exec_cmd('run', utilsCmdArgs='env')?

jasonrahm:
likely bad things.

jasonrahm:
The actual answer is:

"/bin/env: /bin/env: cannot execute binary file "

So I'm guessing I should back off from testing for /bin/bash and just test for starting with /bin/ to catch any other unacceptable arguments passed to bash?

jasonrahm:
@jgruber requested the functionality and really only needs support for bash -c.....

But if there are other important use cases we need to allow, I want to be thorough.

pjbreaux:
I'm for adding the restriction if it will help protect users and notify them of the requirement. And when we get a user that needs to do something without -c, then we can handle it then.

jasonrahm:
Ok, I'll rework the code and resubmit tomorrow. Thanks for the feedback.

jasonrahm:
@caphrim007 this is ready for review.

jasonrahm:
looks like I'm short a test case in coveralls @caphrim007, I'll investigate.

jasonrahm:
ok, @caphrim007, ready for review.

f5/bigip/tm/util/Bash.py
```diff
@@ -0,0 +1,70 @@
...
+from f5.bigip.resource import UnnamedResource
+from f5.utils.util_exceptions import UtilError
+
+
+class Bash(UnnamedResource, CommandExecutionMixin):
+    """BIG-IP® utility command
+
+    .. note::
+
+        This is an unnamed resource so it has not ~Partition~Name pattern
```
caphrim007, line 39: "no" instead of "not". We will fix this in a follow-up PR

test/functional/tm/util/test_bash.py
```diff
@@ -0,0 +1,66 @@
...
+        utilCmdArgs='-c "echo hello >> /var/tmp/test.txt"')
+
+    # commandResult should not be present if this was successful
+    assert 'commandResult' not in bash1.__dict__
+
+    bash2 = mgmt_root.tm.util.bash.exec_cmd(
+        'run',
+        utilCmdArgs='-c df -k')
+
+    # commandResult should b present with data from 'df -k'
```
caphrim007, line 42: "be". Again, we'll fix it in a follow-up PR

caphrim007 (approved):
lgtm

jasonrahm:
thanks, I'll fix those with next command in util.

PR #691, opened 2016-09-13
Adds support for validate_only to the transaction

### LINKED ISSUE #653: add validateOnly support to TransactionContextManager
sometimes you just want to validate that a transaction will be successful, but not actually commit the transaction. This can be done by adding this to the patch payload:

`"validateOnly": true`

caphrim007:
Issues:
Fixes #653

Problem:
The validateOnly functionality of transactions was not exposed to
the end user. Therefore, the user could never know if the transaction
would pass or fail if they ran it without actually running it

Analysis:
This patch adds support to the TransactionContextManager so that
it allows for the validate_only parameter to be provided.

By default, this parameter is set to False to maintain backwards
compatibility.

Tests:
tests/functional/tm/transaction/test_transaction.py::TestTransaction::test_validate_only_true

### DIFF
--- a/f5/bigip/contexts.py
+++ b/f5/bigip/contexts.py
@@ -27,8 +27,33 @@ class TransactionSubmitException(F5SDKError):
 
 
 class TransactionContextManager(object):
-    def __init__(self, transaction):
+    def __init__(self, transaction, validate_only=False):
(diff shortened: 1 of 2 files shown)


jasonrahm:
tests ran as expected and code looks good to me.

Issue #700, opened 2016-09-19
Add ASMFileUploadMixin


wojtek0806:
helper class needed for ASM file-transfer endpoints

wojtek0806:
this has been merged

PR #709, opened 2016-09-19
Adding ASM filemixin

wojtek0806:
#700

Problem:
This is first of a series of PRs aimed at implementing ASM endpoint. This is a first phase of implementation. This PR adds mixin class that will be later used with file-transfer endpoint for ASM.

Analysis:
Added AsmFileTransfer mixin to mixins.py

Tests:
none

### DIFF
--- a/f5/bigip/mixins.py
+++ b/f5/bigip/mixins.py
@@ -44,6 +44,16 @@ class UnsupportedTmosVersion(F5SDKError):
     pass
 
 
+class EmptyContent(F5SDKError):
+    """Raise an error if the returned content size is 0"""
+    pass
+
+
+class MissingHttpHeader(F5SDKError):
+    """We raise this when the expected http header in response is missing"""
+    pass
+
+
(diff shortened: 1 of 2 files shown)


wojtek0806:
@zancas

caphrim007:
Can you elaborate on your "ease of code maintenance" thoughts?

wojtek0806:
if there is need to change/update that part of code relating to ASM file-transfer you can modify the mixin, without affecting other parts of the SDK, I was considering making this mixin inherit from FileUploadMixin, but the whole point of mixins is that they are helper classes they should not inherit from anything than the meta_class

caphrim007:
Do you see a definite discrepancy that you're working around? Or is this just theory right now?

wojtek0806:
no theory, this was already tested, the code is ready for asm, just adding this gradually, and writing unit and functional tests as appropriate

wojtek0806:
have a look for the full code in PR #699

PR #711, opened 2016-09-20
Issue #538 - Adding the qkview utility

jasonrahm:
Updated Files for Qkview
f5/bigip/tm/util/**init**.py

Added Files for Qkview
f5/bigip/tm/util/Qkview.py
f5/bigip/tm/util/test/test_qkview.py
test/functional/tm/util/test_qkview.py


caphrim007:
@jasonrahm there's an issue with the 12.1.0 builds right now, so this will continue to fail. i will figure it out in the AM

caphrim007:
lgtm

PR #722, opened 2016-09-23
Trust Domain - tests failing in 12.1

jasonrahm:
Updated File:
- f5/bigip/tm/cm/trust_domain.py
- test/functional/cm/test_trust_domain.py
  
  Added File:
- f5/bigip/tm/cm/test/test_trust_domain.py

Updates:
- Added unit tests
- Trust Domain never intended to allow create/delete, adding error handlers to prevent their use
- Fixed functional tests given added create/delete restrictions on this resource


jasonrahm:
Note that the create/delete methods will raise an unsupported message for all versions, not just 12.1.

pjbreaux:
You may also want to check out the clustering module in multi_device, as it uses trust_domain create and delete, I believe.

jasonrahm:
ok will do

jasonrahm:
ok, all cluster tests were not impacted, I think the multi-device "trust domain" is not the same as the iControl REST interface cm/trust-domain.

This is ready for review @caphrim007

caphrim007:
lgtm. @pjbreaux any concerns?

Issue #668, opened 2016-08-26
ltm policy functional rewrite in tmos version 12.1 needs attention in sdk


jasonrahm:
In 12.1, policies cannot just be created and attached to virtual servers. There is a workflow to create a draft, then publish, then attach. For existing policies, you need to create a draft of the policy, patch the draft, then publish it. Steps to update a published draft using obscure combo using options and payloads with “command” :-) (provided by Alex A)

```
(venv)ubuntu@ip-172-16-13-92:~/aws-deployments$ curl -u admin:'admin' -H "Content-Type: application/json" -X PATCH -d '{"partition":"Common"}' -sk https://x.x.x.x/mgmt/tm/ltm/policy/~Common~ltm_policy_w_asm_linux_high-Vip1?options=create-draft | python -m json.tool
{
    "controls": [
        "asm"
    ],
    "draftCopy": "/Common/Drafts/ltm_policy_w_asm_linux_high-Vip1",
    "draftCopyReference": {
        "link": "https://localhost/mgmt/tm/ltm/policy/~Common~Drafts~ltm_policy_w_asm_linux_high-Vip1?ver=12.1.0"
    },
    "fullPath": "/Common/ltm_policy_w_asm_linux_high-Vip1",
    "generation": 2697,
    "kind": "tm:ltm:policy:policystate",
    "lastModified": "2016-08-23T21:34:58Z",
    "name": "ltm_policy_w_asm_linux_high-Vip1",
    "partition": "Common",
    "references": {},
    "requires": [
        "http"
    ],
    "rulesReference": {
        "isSubcollection": true,
        "link": "https://localhost/mgmt/tm/ltm/policy/~Common~ltm_policy_w_asm_linux_high-Vip1/rules?ver=12.1.0"
    },
    "selfLink": "https://localhost/mgmt/tm/ltm/policy/~Common~ltm_policy_w_asm_linux_high-Vip1?options=create-draft&ver=12.1.0",
    "status": "published",
    "strategy": "/Common/first-match",
    "strategyReference": {
        "link": "https://localhost/mgmt/tm/ltm/policy-strategy/~Common~first-match?ver=12.1.0"
    }
}

Then you patch the draft copy
[...]

PR #723, opened 2016-09-24
Bugfix.test policy

jasonrahm:


jasonrahm:
@caphrim007 and @pjbreaux this is ready for review. As noted in my commit notes, this doesn't fix the workflow problems introduced in 12. in issue #668, just fixes the test cases temporarily until the workflow is implemented

caphrim007:
lgtm

Issue #719, opened 2016-09-22
Fix GTM tests for 12.1.0


wojtek0806:
GTM tests are failing for 12.1.0, will fix this along with #679

wojtek0806:
Spotted issues with v12 and members subcollection:

On v12.0 getting 404 error if we do not specify partition, but object gets created, for pool members subcollection

On v12.1.0 we are getting 404 error even if we specify partition, but object gets created for pool members subcollection

Need to figure out how we get out of this pickle

wojtek0806:
it is a known issue it seems affecting all of v12.x - ID610441

jasonrahm:
wow, that's particularly ugly. Looks like you might need to overload the create method, and if 404, make another call to check that the object was created, and if so, return the object, otherwise, return the 404.

wojtek0806:
already implemented :) just testing stuff before i set PR

Issue #679, opened 2016-09-09
correct resource classes naming


wojtek0806:
to stay in the set precedent we need to correct Resource class names to be singular not plural, this may/may not apply to Unnamed Resources, opening this issue to look through code and correct where feasible

wojtek0806:
we will have to do this carefully so we do not break backwards compatibility, but suffice to say we should follow the above from now on, i will use this issue to correct Members code for GTM pools

wojtek0806:
@zancas your thoughts? as this was brought up recently and i have not seen anything in our dev guide. Also what would be the impact of us renaming all of the other the classes to follow one convention? This needs to be addressed sooner or later

wojtek0806:
solved

Issue #680, opened 2016-09-12
increase coverage of pool.py in GTM


wojtek0806:
Need to add some more unit tests for members subcollection

PR #727, opened 2016-09-27
Addresses v12.1 changes and incompatibilities in GTM sub-module

wojtek0806:
Fixes Issues: #719,  #679

Updated Files:

```
test/functional/gtm/test_pool.py
test/functional/test_server.py
f5/bigip/tm/gtm/pool.py
f5/bigip/tm/gtm/server.py
```

Updates:

As we started adding Partition to the selflinks in 12.1.0, some tests needed to be amended. There is a serious bug in 12.x branch with members, as described in #719, I have had to create custom create methods for 2 member classes in pools.py.
Corrected subcollection resource names to be singular not plural as per discussion with Tim, amended tests accordingly

### DIFF
--- a/f5/bigip/tm/gtm/pool.py
+++ b/f5/bigip/tm/gtm/pool.py
@@ -32,6 +32,8 @@ from distutils.version import LooseVersion
 from f5.bigip.resource import Collection
 from f5.bigip.resource import OrganizingCollection
 from f5.bigip.resource import Resource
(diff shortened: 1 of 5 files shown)


wojtek0806:
will address the coverage of pool.py in separate PR, tracked under: #680

jasonrahm:
looks like test coverage dropped significantly with the necessary changes.

wojtek0806:
we can live with that for now :)

PR #736, opened 2016-09-27
Issue #730 - Fixes snat functional test for 12.1

jasonrahm:
Updated Files
- test/functional/tm/ltm/test_snat.py
  
  Updates
- Updated bigip fixture to mgmt_root fixture in all test cases
- Changed test assertion from exact match to make sure string was present as the response text has changed slightly in 12.1


jasonrahm:
@caphrim007 snat test is passing now, this is good for review.

Issue #575, opened 2016-07-21
During functional testing against 12.1.0, ltm/policy tests fail with 404s


pjbreaux:
These tests rely upon some default policies to exist which do not in this version. The better fix for tests across all versions would be to have the test (or suite) setup create a policy and test against something newly created. When the suite tears down, the policy would go away. The failure message is captured below for one of the tests:

`
iControlUnexpectedHTTPError: 404 Unexpected Error: Not Found for uri: https://10.190.3.57:443/mgmt/tm/ltm/policy/~Common~_sys_CEC_video_policy
Text: u'{"code":404,"message":"01020036:3: The requested Policy (/Common/_sys_CEC_video_policy) was not found.","errorStack":[],"apiError":3}'
`

PR #741, opened 2016-09-29
Feature.policy drafts

pjbreaux:
@zancas 

This build will break until we release icontrol rest with subpath updates. Will do tomorrow.
#### What's this change do?

Fixes Issues #575 and #668. LTM policies have changed in TMOS 12.1.0, now allowing creation of a draft policy. We need to support this for L7 Content Switching in the agent.
#### Any background context?

This requires a change in icontrol rest as well. To allow subpath to be specified, which is where draft policies are created by default.

Start at the functional tests.


pjbreaux:
Ok, this build should only fail for other 12.1 issues that are not addressed by this PR.

PR #745, opened 2016-09-29
Make dockerfiles respect the requirements.txt pip files

### LINKED ISSUE #744: dockerfiles do not respect requirements.txt
This is a deficiency in docker that it will not be able to detect a change in a file, only a change in the dockerfile. Instead, we need to handle this in the CI process

caphrim007:
Issues:
Fixes #744

Problem:
The dockerfiles maintained a separate list of dependencies and this
would cause issues when those dependencies were updated because they
needed to be updated in two places at once

Analysis:
This patch makes the dockerfiles use the requirements file and behind
the scenes we also adjust the CI build to accomodate this.

Tests:
none

### DIFF
--- a/devtools/dockerfiles/Dockerfile.python2
+++ b/devtools/dockerfiles/Dockerfile.python2
@@ -22,20 +22,14 @@
 FROM python:2
 MAINTAINER @f5networks
 
+COPY . /tmp/src/
+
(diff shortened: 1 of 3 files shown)


pjbreaux:
Looking very good. Thanks for the work.

PR #747, opened 2016-10-01
Feature.util_ciphers

jasonrahm:
Issue #538 - clientssl/serverssl cipher list utility features

Files Updated
- f5/bigip/tm/util/**init**.py
  
  Files Added
- f5/bigip/tm/util/Clientssl_Ciphers.py
- f5/bigip/tm/util/Serverssl_Ciphers.py
- f5/bigip/tm/test/test_clientssl_ciphers.py
- f5/bigip/tm/test/test_serverssl_ciphers.py
- test/functional/tm/util/test_clientssl_ciphers.py
- test/functional/tm/util/test_serverssl_ciphers.py
  
  Notes
- commands are only valid from tmsh in 12.1+
- unit tests for both commands
- functional tests for both commands

### DIFF
--- a/f5/bigip/tm/util/test/test_clientssl_ciphers.py
+++ b/f5/bigip/tm/util/test/test_clientssl_ciphers.py
@@ -16,6 +16,7 @@
 import mock
 import pytest
 
+from distutils.version import LooseVersion
 from f5.bigip import ManagementRoot
 from f5.bigip.tm.util.Clientssl_Ciphers import Clientssl_Ciphers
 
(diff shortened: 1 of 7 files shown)


f5/bigip/tm/util/test/test_clientssl_ciphers.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def test_cssl_exec_missing_args(FakeClientsslCiphers):
+    with pytest.raises(MissingRequiredCommandParameter):
+        FakeClientsslCiphers.exec_cmd('run')
+
+
+def test_command_clientssl_ciphers_wrong_ver(FakeClientsslCiphers):
+    with pytest.raises(UnsupportedTmosVersion):
+        FakeClientsslCiphers.exec_cmd('run', utilCmdArgs='')
```
wojtek0806, line 49: This test fails, as version checking works a bit differently, if you look at this:

https://github.com/F5Networks/f5-common-python/blob/development/f5/bigip/mixins.py#L123

The checking for actual version happens when we instantiate the class as an attribute of container, which in this case would be util, you would have to do something like this to get this working:

https://gist.github.com/wojtek0806/b5c13739969cd4559c866107c38dc004

wojtek0806 (commented):
I see that you fixed the other unit test by applying the fixture for v12, so I wont comment on that line of code.

f5/bigip/tm/util/test/test_clientssl_ciphers.py
jasonrahm: that didn't seem to help any. I'll dig in later tonight, I should have some play time after kids go to bed.

Issue #763, opened 2016-10-05
In order to run register device against older devices its necessary to handle new resources


zancas:
I had to add this patch to support the f5 Boulder lab infrastructure.   It makes the test utility more useful, and has no down-side as fast as I can tell.

PR #764, opened 2016-10-05
Make registration work against older devices.

zancas:
Fixes:  https://github.com/F5Networks/f5-common-python/issues/763


pjbreaux:
And we were originally printing out the resource.raw. Do we want to have that logged on any exception?

zancas:
It's sufficient to log in the new case.  No need to print.

pjbreaux:
Cool. Merging.

PR #776, opened 2016-10-14
ltm node sets 'state' as a read-only attribute

### LINKED ISSUE #775: ltm node sets 'state' as a read-only attribute
The 'state' of a node can be modified through REST and is not strictly read-only. We can remove this and test to ensure it works as expected. The test will consist of a user creating a node with a specific state, updating to a specific state, then modifying the state.

pjbreaux:
Issues:
Fixes #775

Problem:
The 'state' of a node can be modified through REST and is not strictly
read-only. We can remove this and test to ensure it works as expected.
The test will consist of a user creating a node with a specific state,
updating to a specific state, then modifying the state.

Analysis:
Remove 'state' as read-only attribute, then ensured a call to update or
modify does the right thing if state is present and it has a value of
unchecked.

Tests:
Wrote new functional tests to check this behavior

### DIFF
--- a/f5/bigip/tm/ltm/node.py
+++ b/f5/bigip/tm/ltm/node.py
@@ -29,6 +29,7 @@ REST Kind
 
 from f5.bigip.resource import Collection
 from f5.bigip.resource import Resource
+from f5.sdk_exception import NodeStateModifyUnsupported
 
 
 class Nodes(Collection):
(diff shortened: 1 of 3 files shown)


pjbreaux:
I did manually test that you couldn't change the state to 'user-down' then back to 'unchecked'. Still get in iControlUnexpectedHTTPError.

PR #785, opened 2016-10-24
Adding Token Auth to SDK

### LINKED ISSUE #487: iControlRESTSession needs to support the token=true argument for token auth
Starting with version 12.0 of TMOS token based authentication and authorization is supported by iControl REST.

BIG-IP remote-users access via iControl REST requires token based authorization. f5-common-python access to BIG-IP needs to support token authorization when access BIG-IP devices which are configured for remote-user support. This is a base requirement for all MSPs and most enterprises.

Use of token based authorization needs to be an option parameter for the connection as support for TMOS versions prior to 12.0 is also a requirement.

wojtek0806:
Issues:
Fixes #487

Problem:
SDK does not support token auth, so it will not be usable with situations where remote auth is required. The underlying iControlRest library has this functionality already.

Analysis:
Added a kwarg 'token' option, with the default set to False.

Tests:
Flake 8

### DIFF
--- a/f5/bigip/__init__.py
+++ b/f5/bigip/__init__.py
@@ -43,10 +43,12 @@ class ManagementRoot(PathElement):
         timeout = kwargs.pop('timeout', 30)
         port = kwargs.pop('port', 443)
         icontrol_version = kwargs.pop('icontrol_version', '')
+        token = kwargs.pop('token', False)
         if kwargs:
             raise TypeError('Unexpected **kwargs: %r' % kwargs)
         # _meta_data variable values
-        iCRS = iControlRESTSession(username, password, timeout=timeout)
+        iCRS = iControlRESTSession(username, password, timeout=timeout,
+                                   token=token)
         # define _meta_data
         self._meta_data = {
             'allowed_lazy_attributes': [Tm, Cm, Shared],


wojtek0806:
@caphrim007 , you requested this :)

jasonrahm:
That supports passing the token, but what about generating it?

wojtek0806:
isn't that the job for logonprovider?

jasonrahm:
where is that code?

wojtek0806:
that is already inside iControl library we use for this:

https://github.com/F5Networks/f5-icontrol-rest-python/blob/1.0/icontrol/authtoken.py

jasonrahm:
functionality looks good on my end with tacacs but are you going to add tests? Looks like you could add to f5/bigip/test/unit/test__init__.py a FakeBigIPWithToken fixture to work with at a minimum if external auth won't be available for functional tests.

wojtek0806:
there is nothing to test here on unit test side, other than asserting that we do not throw up exception about invalid kwargs something like this:

https://github.com/F5Networks/f5-common-python/blob/development/f5/bigip/test/functional/test__init__.py#L20

https://github.com/F5Networks/f5-common-python/blob/development/f5/bigip/test/unit/test___init__.py#L76

But that already covers that part of code. Since most of the magic happens behind the scenes, the only thing one can reliably do is to for example create/load example object and check for certain attributes that would be telling if we are using tokens or not. That I could do..but the question is do we want it :) ?

jasonrahm:
I get a 401 error if I try to use my remote auth credentials with basic auth, so that is testable. And the simple assertion in the unit tests should be fine. But that's my $.02. What say you @caphrim007?

wojtek0806:
You need to have this set in Jenkins to be able to test as functional test

caphrim007:
@jasonrahm is correct that we need tests. @wojtek0806 is correct that jenkins does not have this harness in it. so we'll track the jenkins issue here https://github.com/F5Networks/f5-common-python/issues/786

PR #789, opened 2016-10-24
Add functional tests for LTM policy deployments

### LINKED ISSUE #788: Add functional tests for LTM policy deployments
When deploying a policy with the agent, we would like to deploy the policy along with the rules and conditions and actions in one POST or PUT. We need a test to assure this operation produces the expected state on the device.

pjbreaux:
@zancas 

Issues:
Fixes #788 and #790 

Problem:
When deploying a policy with the agent, we would like to deploy the
policy along with the rules and conditions and actions in one POST or
PUT. We need a test to assure this operation produces the expected state
on the device.

Analysis:
Added two new test to check this behavior. Found that we can do a PUT on
a policy to change the state of it and the accompanying rules
immediately. We validate that the state is as expected immediately after
updating.

Tests:
Both new tests pass against 12.1 and 11.6

### DIFF
--- a/f5/bigip/tm/ltm/policy.py
+++ b/f5/bigip/tm/ltm/policy.py
@@ -44,6 +44,11 @@ class OperationNotSupportedOnPublishedPolicy(F5SDKError):
     pass
 
 
+class NonExtantPolicyRule(F5SDKError):
+    '''Raise if a rule does not exist on the device.'''
+    pass
+
+
 class Policys(Collection):
     """BIG-IP® LTM policy collection."""
     def __init__(self, ltm):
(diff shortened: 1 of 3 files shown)


pjbreaux:
@zancas, the issue with all rules in 11.6.0 (whether existing or not) returning 200 OK has been fixed.

pjbreaux:
@caphrim007: Let me know if you have the bandwidth to review this. It is a fix for a couple of issues. I've overridden exists and load for a rule, so it doesn't just return 200 OK even if the rule doesn't exist. That problem is only evident in 11.6.0.

caphrim007:
this looks fine.

PR #802, opened 2016-11-10
Create collection and resource under virtual to allow access to policies

### LINKED ISSUE #800: Create collection and resource under virtual to allow access to policies
The virtual resource currently does not have SDK access to its policies. Therefore we cannot perform CRUD operations on these policies. We need to add the collection for Policies_s and the resource Policies. The collection is shown below:

{
  "kind": "tm:ltm:virtual:policies:policiescollectionstate",
  "selfLink": "https://localhost/mgmt/tm/ltm/virtual/~Common~test/policies?ver=12.1.0",
  "items": []
}

This will require some refactor work in the Policy module, mostly surrounding publish and legacy keywords. We will strive to preserve the old behavior and we will attempt to do the right thing when these types of arguments are given for bigip versions that are less than 12.1.0.

pjbreaux:
@caphrim007 

Issues:
Fixes #800

Problem:
The virtual resource currently does not have SDK access to its policies.
Therefore we cannot perform CRUD operations on these policies. We need
to add the collection for Policies_s and the resource Policies.

Analysis:
I had to refactor the policy _create method to make the logic read
better from a developer's viewpoint. We changed some of the underlying
implementation and we have loosened restrictions for behaviour against 
older tmos versions. And in this respect, we have broken backwards compatibility.

Tests:
New tests for virtual, and refactored tests for policy all pass


pjbreaux:
@caphrim007: wait until @jlongstaf gives his go-ahead on a few tests he's running before you look at this

pjbreaux:
Also @caphrim007: this breaks backwards compatibility because I changed source code and broke a test, then I had to go fix up the test to get it to work. This is our tell-tale sign that we have broken backwards compatibility. We just need to be aware of this when we make our next release, which will probably be tomorrow.

jlongstaf:
@caphrim007, @pjbreaux : these changes work fine for me. Please feel free to merge.

caphrim007:
@pjbreaux so is this text, "but we preserved existing functionality", in the description not true?

pjbreaux:
Tis a lie! I tried to preserve the old functionality, but couldn't, because the old way kind of sucked.

PR #804, opened 2016-11-11
Fix issue of existence for policies under virtual servers

pjbreaux:
@caphrim007 

#### What's this change do?
I commonized the logic to check for existence of a policy under a virtual to be used by other resources that need it. It is now in the mixins module.

#### Any background context?
This issue has been encountered before. Here, if we check for existence or attempt to load a missing policy from a virtual server's namespace, it will always succeed when using TMOS 11.6.0. Meaning that the GET call returns a 200 OK, whether the policy exists or not. We need to iterate of the collection of items from the container and check if the object we need exists by name.

Start at the mixin and then move into the code chances in each resource.


pjbreaux:
@caphrim007: okay, I think this is ready for review. Thanks for your assistance!

PR #808, opened 2016-11-14
Adds remaining parameter parsing for iapps

### LINKED ISSUE #807: factor other supported parameters and formats into iapp_parser
The iapp_parser util supports the common formats for iapps but does not support all of them, particularly those that are part of the iapp templates that F5 releases on downloads.f5.com

caphrim007:
Issues:
Fixes #807

Problem:
The iapp_parser does a good job of parsing the common formats of
iapps but leaves out some of the more edge cases.

Analysis:
This patch fixes the iapp parser to take into account the remaining
things that can be parsed and reverses some of the expectations
the iapp parser made that were incorrect.

For example, the assumption that requires-modules cannot have a
value of "{ }"

This patch changes the tests that existed because some of those
existing tests were incorrect. Does this mean a backwards incompatible
change?

Tests:
unit tests

### DIFF
--- a/f5/utils/iapp_parser.py
+++ b/f5/utils/iapp_parser.py
@@ -27,16 +27,43 @@ class IappParser(object):
         u'role-acl'
     ]
 
-    tcl_list_for_attr_re = '{(\s*\w+\s*)+}'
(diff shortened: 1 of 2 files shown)


caphrim007:
@pjbreaux here are the changes I mentioned today

PR #809, opened 2016-11-15
Fixes Update method #807

### LINKED ISSUE #807: factor other supported parameters and formats into iapp_parser

wojtek0806:
Issues:
Fixes #807

Problem:
The update method strips _meta_data during its update process. This works well when the update suceeds. The problem occurs when we receive exception from bigip. Not only we fail to update (for one reason or the other which is what we intended) but we also leave the calling object with this _meta_data removed.

Analysis:
This patch added the logic to return the object to its original state as well as passing on the raised exception. This means that we do see failure as intended but we no longer have objects with _meta_data stripped.

Tests:
unit tests
flake 8

### DIFF
--- a/f5/bigip/resource.py
+++ b/f5/bigip/resource.py
@@ -100,8 +100,8 @@ from f5.bigip.mixins import LazyAttributeMixin
 from f5.bigip.mixins import ToDictMixin
 from f5.sdk_exception import F5SDKError
 from f5.sdk_exception import UnsupportedMethod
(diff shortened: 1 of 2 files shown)


wojtek0806:
@pjbreaux or @caphrim007 whenever it finishes testing please merge :)

PR #813, opened 2016-11-15
Fixes default success assertions in util tests

### LINKED ISSUE #812: util tests are defaulting to "passed"
The assertions made in the functional tests for the util endpoints are defaulting to "passed" because their assertion has been placed inside of the `with ... exception` block.

This causes the exception to be raised and caught by pytest, but since the assertion is coded inside of the exception block (after the exception is raised) the assertion never happens and the test succeeds by default

caphrim007:
Issues:
Fixes #812

Problem:
The tests for all of the util functional tests were always returning
`true` because the assertions were being skipped

Analysis:
this patch fixes the tests by moving the assertion up and also refactoring
the tests so that they are in their own test functions and classes where
appropriate

Tests:
functional

### DIFF
--- a/f5/bigip/tm/util/test/functional/test_bash.py
+++ b/f5/bigip/tm/util/test/functional/test_bash.py
@@ -1,4 +1,3 @@
-
 # Copyright 2016 F5 Networks Inc.
 #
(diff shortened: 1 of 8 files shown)


wojtek0806:
lgtm

PR #815, opened 2016-11-16
Fixes test file name

### LINKED ISSUE #814: correct test filename for sys/application endpoint
the filename does not match the file being tested. Change this so that we can isolate jenkins tests

caphrim007:
Issues:
Fixes #814

Problem:
the filename for testing the sys/application endpoint does not match
the name of the file it is testing

Analysis:
this patch corrects the filename so that test isolation will work

Tests:
functional


caphrim007:
@wojtek0806

PR #830, opened 2016-11-22
LTM policy needs to be capable of update and modify with 12.1.0 with legacy

### LINKED ISSUE #829: LTM policy needs to be capable of update and modify with 12.1.0 with legacy
There is a legacy key for deploying and modifying policies in 12.1.0. When set to True, this deploys or modifies a policy as if it were one on tmos versions prior to 12.1.0. We need to change the modify and update operations to expect legacy as a keyword argument and allow the user to update as they see fit.

pjbreaux:
@jlongstaf 

Issues:
Fixes #829

Problem:
There is a legacy key for deploying and modifying policies in 12.1.0.
When set to True, this deploys or modifies a policy as if it were one on
tmos versions prior to 12.1.0. We need to change the modify and update
operations to expect legacy as a keyword argument and allow the user to
update as they see fit.

Analysis:
Made changes to modify and update to allow legacy to be passed in and
used for changes against a 12.1.0 machine. If legacy is passed in and
the tmos version is < 12.1.0, it is simply ignored.

Tests:
All tests against 12.1.0 and 11.6.0 for policies pass

### DIFF
--- a/f5/bigip/tm/ltm/policy.py
+++ b/f5/bigip/tm/ltm/policy.py
@@ -101,9 +101,11 @@ class Policy(Resource):
         :raises: OperationNotSupportedOnPublishedPolicy
         '''
 
+        legacy = patch.pop('legacy', False)
         tmos_ver = self._meta_data['bigip']._meta_data['tmos_version']
         if 'Drafts' not in self._meta_data['uri'] and \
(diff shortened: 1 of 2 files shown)


jlongstaf:
Looks good. Please go ahead and merge.

wojtek0806:
merging

PR #834, opened 2016-11-26
What: Adding token authentication option to conftest.py

jasonrahm:
Why: So we can start testing our functional tests against token auth to find the disparity between local and remote auth credentials

Only file changed: conftest.py


jasonrahm:
ready for review @caphrim007

Issue #835, opened 2016-11-28
DeviceProvidesIncompatibleKey exception when 'not' is given as object attribute


pjbreaux:
For LTM Conditions, 'not' is a valid JSON field, and this is a valid Python keyword. We can get around it (which is what we are doing currently) by doing a setattr and getattr, but this is not sustainable. We must brainstorm a few ideas to convert such things that can be accessible via Python dotted attribute access. (condition._not or condition.invert).

pjbreaux:
Implemented a helper function where python keywords are appended with a trailing underscore. This change takes care of the outgoing request and the reverse translation on the response.

Issue #836, opened 2016-11-29
'not' attribute cannot be set to False once it is set to True on an LTM condition


pjbreaux:
The 'not' attribute can only be set to True for an LTM condition. When trying to set it back to False, which is the default value, the device does not honor the request. I have tried this on tmos versions: 11.6.1 and 12.1.1. I wrote a test for issue #835, but it is currently skipped.

```
{
"kind": "tm:ltm:policy:rules:conditions:conditionsstate",
"name": "0",
"fullPath": "0",
"generation": 16,
"selfLink": "https://localhost/mgmt/tm/ltm/policy/~Common~Drafts~racetest/rules/test_rule/conditions/0?ver=12.1.1",
"all": true,
"caseInsensitive": true,
"equals": true,
"external": true,
"httpHeader": true,
"index": 0,
"tmName": "TESTHEAD",
"not": true,
"present": true,
"remote": true,
"request": true,
"values": [
  "yes"
],
}
```

wojtek0806:
this was merged already, closing

PR #837, opened 2016-11-29
DeviceProvidesIncompatibleKey exception when 'not' is given as object attribute

pjbreaux:
@caphrim007 @jlongstaf 

Fixes: Issue #835 

#### Any background context?
For LTM Conditions, 'not' is a valid JSON field, and this is a valid
Python keyword. We can get around it (which is what we are doing
currently) by doing a setattr and getattr, but this is not sustainable.
We must brainstorm a few ideas to convert such things that can be
accessible via Python dotted attribute access. (condition._not or
condition.invert).

#### What's this change do?
Implemented a helper function where python keywords are appended with a
trailing underscore. This change takes care of the outgoing request and
the reverse translation on the response. Implemented a few new tests, but unconvered a new bug #836, which highlights an issue on the device around setting the 'not' attribute.

This is a fix needed for running tests for L7 Content Switching in the OpenStack agent and driver.

### DIFF
--- a/f5/bigip/tm/gtm/test/functional/test_rule.py
+++ b/f5/bigip/tm/gtm/test/functional/test_rule.py
@@ -96,6 +96,22 @@ class TestCreate(object):
         )
         assert 'check syntax' in rule1.apiAnonymous
 
+    def test_create_optional_args_bad_order(self, request, mgmt_root):
+        '''The device requires the check key to be before apiAnonymous.
+
+        Apparently order matters against the device when it comes to
+        the check key value and the apiAnonymous key value in the JSON blob.
+        This test is to ensure we at least try for a different ordering.
+        '''
+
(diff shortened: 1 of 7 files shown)


f5/bigip/tm/gtm/test/functional/test_rule.py
```diff
@@ -28,10 +28,7 @@
 )
 
 
-RULE = '''when LB_SELECTED {
-   set wipHost [LB::server addr]
-}
-'''
+RULE = '''when LB_SELECTED {set wipHost [LB::server addr]}'''
```
pjbreaux: This was just a test change to see if the newlines had an bearing on the failure. They don't.

jlongstaf (approved):
Looks fine.

caphrim007:
that we have to do this hack at all makes me a sad panda, but lgtm

PR #839, opened 2016-11-29
Corrects tox targets

### LINKED ISSUE #838: tox targets missing py27 and including invalid bigip version
these should be fixed

caphrim007:
Issues:
Fixes #838

Problem:
The tox targets did not include python 2.7 and included a version of
BIG-IP that did not xist (11.5.1)

Analysis:
This patch correct the bigip version and adds python 2.7 support

Tests:


caphrim007:
@pjbreaux

PR #912, opened 2017-01-12
Mv code from os test

zancas:
@pjbreaux 
Fixes: #911 

Here's another change that will allow us to decommission `f5_openstack_test`.


pjbreaux:
Merging.

PR #916, opened 2017-01-13
update install requires to cap deps

zancas:


zancas:
@caphrim007 I think this can go in.

PR #931, opened 2017-01-18
Exceptions move final

wojtek0806:
Problem:
Currently exception classes are scattered all over SDK, this will lead to unecessary duplication and confusion

Analysis:
Final Exceptions Move PR

Tests:
Flake8


wojtek0806:
we need that 1.3 interface so we can run those tests @caphrim007

wojtek0806:
it fails due to jenkins limitation, merging

PR #945, opened 2017-01-23
Makes the name parameter a required load parameter for the user object

caphrim007:
Issues:

Problem:
User module did not have correct required_load_parameter specified

Analysis:
this fixes it

Tests:


jasonrahm:
lgtm, will merge when tests complete.

PR #954, opened 2017-01-25
Feature.sysprovision

jgruber:
The test has to be run on at least LTM qcow2 images setup with the appropriate flavor support from the VE support matrix. You can not make provisioning changes to 1SLOT images. They will return '400 errors - Memory limit exceed' if the VE instances are not setup with at least 2104 MB dedicated to TMM.

### DIFF
--- a/f5/bigip/tm/sys/test/functional/test_provision.py
+++ b/f5/bigip/tm/sys/test/functional/test_provision.py
@@ -15,9 +15,9 @@
 
 
 class TestProvision(object):
-    def test_Provision(self, request, bigip):
+    def test_Provision(self, request, mgmt_root):
         # Load
-        ltmprov = bigip.sys.provision.ltm.load()
+        ltmprov = mgmt_root.sys.provision.ltm.load()
         assert ltmprov.level == 'nominal'
         assert ltmprov.name == 'ltm'
 
--- a/f5/bigip/tm/sys/test/functional/test_provision.py
+++ b/f5/bigip/tm/sys/test/functional/test_provision.py
@@ -1,0 +1,31 @@
+# Copyright 2016 F5 Networks Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


f5/bigip/tm/sys/test/functional/test_provision.py
```diff
@@ -17,7 +17,7 @@
 class TestProvision(object):
     def test_Provision(self, request, bigip):
         # Load
-        ltmprov = bigip.sys.provision.ltm.load()
+        ltmprov = bigip.tm.sys.provision.ltm.load()
```
wojtek0806: Bigip fixture is deprecated and it is better to use mgmt_root, with that said we I believe we have changed the fixture so it is using mgmt_root as its argument:

https://github.com/F5Networks/f5-common-python/blob/development/f5sdk_plugins/fixtures.py#L102

so you can remove the TM.

wojtek0806:
@jgruber 
Is there a way to detect if we have 1 SLOT VE beforehand? Cause if it is then it would be useful to write a helper function that would simply not allow provisioning endpoint to be run

caphrim007:
@jgruber if you can fix up the issue @wojtek0806 mentioned, i'll review and merge

wojtek0806:
@jgruber I agree with @jasonrahm opinion, that this could use some more functional tests, also some basic unit testing would be great:

````

rahmen.empire [3:34 AM] 
looking a @jgruberf5's PR. We need to add unit/functional tests for code coverage, and also, given how long provisioning takes, to do so we'll need to use the asynchronous tasks added in 12.0. An async task changes the URL from /mgmt/tm/sys/provision to /mgmt/tm/task/sys/provision, and then you need to check taskState for COMPLETED before moving on to the next async task.

[3:35]  
so would we need to build out something similar to transactions?

````

caphrim007:
adding unit and functional tests in a separate PR as this functionality is required now for the ansible modules

wojtek0806:
@caphrim007 was this not addressed with #1270 ?

PR #971, opened 2017-02-08
Adds properties method 

wojtek0806:
this is just a draft, and probably will need to have it changed slightly for python3. Tell me if this is what you are looking for @caphrim007

### DIFF
--- a/f5/bigip/resource.py
+++ b/f5/bigip/resource.py
@@ -687,6 +687,12 @@ class ResourceBase(PathElement, ToDictMixin):
             config_dict.pop(key2)
         return config_dict
 
+    @property
+    def no_meta(self):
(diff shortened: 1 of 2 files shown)


caphrim007:
All I was doing in the meantime was running to_dict() and then doing a pop('_meta_data', None). I think the function needs a name like attrs or something like that. no_meta, while true, is also not necessarily the purpose of the method.

The use case is that I was to get all of the things that would be considered attributes of the resource if I were to do a normal curl of the endpoint.

wojtek0806:
So you like the solution? 

For name of method maybe properties?

caphrim007:
I think properties is fine

wojtek0806:
this will fail for 2.6 python as we use dict comprehension. Since we are about to drop python 2.6 support from Travis, I am going to merge this regardless

Issue #968, opened 2017-02-03
Adding LTM policy to a virtual server produces 404, but succceeds anyway


pjbreaux:
Most of the content of this issue is captured in the agent bug found [here](https://github.com/F5Networks/f5-openstack-lbaasv2-driver/issues/420). This failure seems to occur on BIG-IP 11.5.4, but the fix here will protect other BIG-IP versions as well.

When doing a POST to add a policy to a virtual server, the policy is actually added, but REST returns a 404.

```
{
"code": 404,
"message": "Object not found - /Project_0b6a552bae994078b3b3754b61747abe/test_policy",
"errorStack": [],
}
```

In addition to this, when trying to GET that policy after it has been added, this also fails. We need a workaround in the sdk, and we need new tests to check this behavior.

wojtek0806:
duplicate #833 ?

pjbreaux:
@wojtek0806: hah, good catch. I don't even remember last week half the time. Thanks.

wojtek0806:
will keep this one open until we implement fix for #833

wojtek0806:
11.5.4:

- We cannot get collection as the 'items' is nested inside policiesReference:
````
{
  "kind": "tm:ltm:virtual:policies:policiescollectionstate",
  "selfLink": "https://localhost/mgmt/tm/ltm/virtual/~Common~testy/policies?ver=11.5.4",
  "policiesReference": {
    "link": "https://localhost/mgmt/tm/ltm/virtual/~Common~testy/policies?ver=11.5.4",
    "isSubcollection": true,
    "items": [
      {
        "kind": "tm:ltm:virtual:policies:policiesstate",
        "name": "bugtest",
        "partition": "Common",
        "fullPath": "/Common/bugtest",
        "generation": 102,
        "selfLink": "https://localhost/mgmt/tm/ltm/virtual/~Common~testy/policies/~Common~bugtest?ver=11.5.4"
      }
    ]
  }
}
````
- GET to the selfLink results in 404

Fix for this should more or less follow this logic:

- Overwrite get_collection() method to handle the above peculiarity
- Create the subcollection as normal
- Intercept the 404 error that comes back
- Since GET to selfLink also breaks, issue get_collection() and iterate through list of 'items' 
looking for the 'name' attribute that would match the object we created. This is necessary as we will probably have more than item in there
- Modify load() method for virtual.policys_s.policys object so it utilizes get_collection() mechanism as described above.

[...]

Issue #833, opened 2016-11-23
Adding a policy to a virtual on 12.1.1 produces 404, but succeeds in adding anyway


pjbreaux:
When adding an LTM policy to a virtual server on TMOS version BIG-IP 12.1.1 Build 2.0.204 Hotfix HF2, a 404 error is returned to the client, but the policy is added anyway. I successfully reproduced this against an 12.1.1 box with Postman.

```POST to https://10.190.7.221/mgmt/tm/ltm/virtual/~Project_e43b44cb2a22431f8c0de68e9349fc11~Project_d3447d86-7d57-47fe-bd7e-cea44c390316/policies```

with JSON:

```{"partition": "Project_e43b44cb2a22431f8c0de68e9349fc11", "name": "wrapper_policy"}```

Return is:

```
{
  "code": 404,
  "message": "Object not found - wrapper_policy",
  "errorStack": [],
  "apiError": 1
}
```
```GET to https://10.190.7.221/mgmt/tm/ltm/virtual/~Project_e43b44cb2a22431f8c0de68e9349fc11~Project_d3447d86-7d57-47fe-bd7e-cea44c390316/policies```

Returns:
```
{
  "kind": "tm:ltm:virtual:policies:policiescollectionstate",
  "selfLink": "https://localhost/mgmt/tm/ltm/virtual/~Project_e43b44cb2a22431f8c0de68e9349fc11~Project_d3447d86-7d57-47fe-bd7e-cea44c390316/policies?ver=12.1.1",
  "items": [
    {
      "kind": "tm:ltm:virtual:policies:policiesstate",
      "name": "wrapper_policy",
      "partition": "Project_e43b44cb2a22431f8c0de68e9349fc11",
      "fullPath": "/Project_e43b44cb2a22431f8c0de68e9349fc11/wrapper_policy",
      "generation": 2069,
      "selfLink": "https://localhost/mgmt/tm/ltm/virtual/~Project_e43b44cb2a22431f8c0de68e9349fc11~Project_d3447d86-7d57-47fe-bd7e-cea44c390316/policies/~Project_e43b44cb2a22431f8c0de68e9349fc11~wrapper_policy?ver=12.1.1"
    }
  ]
}
```

wojtek0806:
why am i not surprised :) found this issue on two endpoints in GTM. Feel free to look at create method in Gtm Pools

wojtek0806:
Unless this is a regression it should work.

[...]

PR #978, opened 2017-02-12
Fixes #975 #968 #833

### LINKED ISSUE #975: ltm virtual tests need refactoring
I will refactor these tests so they are tearing down the created resources properly. As currently we are utilizing device_snapshot_fixture, and the tests need to be able to cleanup after they finish, rather than rely on it.

wojtek0806:
Fixes #975 #968 #833

Problem:
There multiple issues with creating/loading with virtual server policies subcollection in 11.5.4 and 12.1.1. Some of the virtual server tests required some refactor and supplementation.

Analysis:
Added custom _create, _load methods to the resources to work around issues in 11.5.4 and 12.1.1. Refactored some of the virtual server functional tests

Tests:
Functional
Flake8

### DIFF
--- a/f5/bigip/mixins.py
+++ b/f5/bigip/mixins.py
@@ -401,3 +401,10 @@ class CheckExistenceMixin(object):
             if item.name == item_name:
                 return True
         return False
+
+    def _return_object(self, container, item_name):
+        """Helper method to retrieve the object"""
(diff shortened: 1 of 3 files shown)


wojtek0806:
ignoring py2.6 error

PR #984, opened 2017-02-15
Feature.mgmtint

jasonrahm:
updated test cases to reflect jenkins environment

### DIFF
--- a/f5/bigip/tm/sys/management_ip.py
+++ b/f5/bigip/tm/sys/management_ip.py
@@ -1,0 +1,62 @@
+# coding=utf-8
+#
+# Copyright 2016 F5 Networks Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+
+"""BIG-IP® system syslog module
+
+REST URI
+    ``http://localhost/mgmt/tm/sys/management-ip``
+
+GUI Path
+    ``System --> Platform``
+
+REST Kind
+    ``tm:sys:management-ip:*``
+"""
+
+from f5.bigip.resource import Collection
+from f5.bigip.resource import Resource
+from f5.sdk_exception import UnsupportedMethod
+
+
+class Management_Ips(Collection):
+    """BIG-IP® management-ip collection"""
+    def __init__(self, sys):
+        super(Management_Ips, self).__init__(sys)
+        self._meta_data['allowed_lazy_attributes'] = [Management_Ip]
+        self._meta_data['attribute_registry'] =\
+            {'tm:sys:management-ip:management-ipstate': Management_Ip}
+
+
+class Management_Ip(Resource):
+    """BIG-IP® management-ip resource"""
+    def __init__(self, Management_Ips):
+        super(Management_Ip, self).__init__(Management_Ips)
+        self._meta_data['required_json_kind'] = \
+            'tm:sys:management-ip:management-ipstate'
+        self._meta_data['required_creation_parameters'].update(('name',))
+
+    def load(self, **kwargs):
+        kwargs['transform_name'] = True
+        return self._load(**kwargs)
+
+    def modify(self, **kwargs):
+        raise UnsupportedMethod(
+            "%s does not support the modify method" % self.__class__.__name__)
+
+    def update(self, **kwargs):
+        raise UnsupportedMethod(
+            "%s does not support the update method" % self.__class__.__name__)
--- a/f5/bigip/tm/sys/management_ip.py
+++ b/f5/bigip/tm/sys/management_ip.py
@@ -29,7 +29,6 @@ REST Kind
 
 from f5.bigip.resource import Collection
 from f5.bigip.resource import Resource
-from f5.sdk_exception import UnsupportedMethod
 
 
 class Management_Ips(Collection):
@@ -47,16 +46,7 @@ class Management_Ip(Resource):
         super(Management_Ip, self).__init__(Management_Ips)
         self._meta_data['required_json_kind'] = \
             'tm:sys:management-ip:management-ipstate'
-        self._meta_data['required_creation_parameters'].update(('name',))
 
     def load(self, **kwargs):
         kwargs['transform_name'] = True
         return self._load(**kwargs)
-
-    def modify(self, **kwargs):
-        raise UnsupportedMethod(
-            "%s does not support the modify method" % self.__class__.__name__)
-
-    def update(self, **kwargs):
-        raise UnsupportedMethod(
-            "%s does not support the update method" % self.__class__.__name__)
--- a/f5/bigip/tm/sys/management_ip.py
+++ b/f5/bigip/tm/sys/management_ip.py
@@ -1,6 +1,6 @@
 # coding=utf-8
 #
-# Copyright 2016 F5 Networks Inc.
+# Copyright 2017 F5 Networks Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


jasonrahm:
@wojtek0806 or @caphrim007 this is ready for review. Note that the create method doesn't fully test a working create and the delete method is not tested to protect the management interface connectivity for all other tests. That logic is there, just commented out.

f5/bigip/tm/sys/management_ip.py
```diff
@@ -0,0 +1,62 @@
+# coding=utf-8
+#
+# Copyright 2016 F5 Networks Inc.
```
caphrim007: we're now in 2017

f5/bigip/tm/sys/management_route.py
```diff
@@ -0,0 +1,50 @@
+# coding=utf-8
+#
+# Copyright 2016 F5 Networks Inc.
```
caphrim007: year

f5/bigip/tm/sys/test/functional/test_management_ip.py
```diff
@@ -0,0 +1,109 @@
+# Copyright 2015-2106 F5 Networks Inc.
```
caphrim007: same year. also, needs the python shebang at the top

caphrim007:
@jasonrahm can you really not update the management ip through the api?

f5/bigip/tm/sys/management_route.py
jasonrahm: heh, I suppose I should read comment fields I copy over from other files more carefully...I can clean up here or I can submit a PR to update all source files for copyright info.

caphrim007: we dont need to update all source files in the source tree for copyright, just the new files that we add.

jasonrahm:
You can update the management IP but it is done through create, not modify or update.

caphrim007:
so you can create and you can delete, and updating is no different from creating?

jasonrahm:
A POST will overwrite existing management IP as long as it is different than the existing one. PUT/PATCH will not do anything to the management IP.

caphrim007:
I'm wondering then if, for consistency and user experience's sake, if we should just alias modify commands to create, and hide that implementation detail from the user

f5/bigip/tm/sys/test/functional/test_management_ip.py
jasonrahm: when did we start adding the shebang to files? I haven't seen that in the other modules

jasonrahm:
do you mean that we should only support .update() or .modify() , but in those, call the create method instead?

caphrim007:
Ignore my shebang comment. You're correct.

For modify and update, yes, if it behaves where the experience of "updating" is essentially just calling create, then i would think the user would think they are updating, when in dact we're just calling create on their behalf

jasonrahm:
ok i will refactor, thanks

src: mobile

On Feb 15, 2017, at 7:28 PM, Tim Rupp <notifications@github.com<mailto:notifications@github.com>> wrote:


Ignore my shebang comment. You're correct.

For modify and update, yes, if it behaves where the experience of "updating" is essentially just calling create, then i would think the user would think they are updating, when in dact we're just calling create on their behalf

-

wojtek0806:
what happens if you don't have mgmt IP set?, i think it would be better if we leave create(), and just redo modify() or update() or both to use POST. This way this will be cleaner than aliasing and doing all weird voodoo stuff

@jasonrahm

jasonrahm:
yes, create needs to still be there for initial setting of management interface if absent.

wojtek0806:
just override _update() method and make_modify() not allowed

f5/bigip/tm/sys/management_ip.py
```diff
@@ -0,0 +1,62 @@
...
+            {'tm:sys:management-ip:management-ipstate': Management_Ip}
+
+
+class Management_Ip(Resource):
+    """BIG-IP® management-ip resource"""
+    def __init__(self, Management_Ips):
+        super(Management_Ip, self).__init__(Management_Ips)
+        self._meta_data['required_json_kind'] = \
+            'tm:sys:management-ip:management-ipstate'
+        self._meta_data['required_creation_parameters'].update(('name',))
```
wojtek0806: that should be removed as we already have this as default in Resource super class

f5/bigip/tm/sys/management_route.py
```diff
@@ -0,0 +1,50 @@
...
+        self._meta_data['attribute_registry'] =\
+            {'tm:sys:management-route:management-routestate': Management_Route}
+
+
+class Management_Route(Resource):
+    """BIG-IP® system management-route resource"""
+    def __init__(self, Management_Routes):
+        super(Management_Route, self).__init__(Management_Routes)
+        self._meta_data['required_creation_parameters'].update(
+            ('name', 'network', 'gateway'))
```
wojtek0806: again you can leave out the 'name' parameter as this is already there.  Should be something like this instead:

` self._meta_data['required_creation_parameters'].update(('network', 'gateway'))`

f5/bigip/tm/sys/management_ip.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def load(self, **kwargs):
+        kwargs['transform_name'] = True
+        return self._load(**kwargs)
+
+    def modify(self, **kwargs):
```
wojtek0806: not entirely true, there is a property called 'description', and 'app-service' which can be manipulated, so this should be allowed

f5/bigip/tm/sys/management_ip.py
```diff
@@ -0,0 +1,62 @@
...
+        raise UnsupportedMethod(
+            "%s does not support the modify method" % self.__class__.__name__)
+
+    def update(self, **kwargs):
```
wojtek0806: not entirely true, there is a property called 'description', and 'app-service' which can be manipulated, so this should be allowed

f5/bigip/tm/sys/test/functional/test_management_ip.py
```diff
@@ -0,0 +1,109 @@
...
+#         mgmt_root.tm.sys.management_ips.management_ip.load(
+#             name=mip[0].name)
+#     assert ') was not found.' in str(err.value.message)
+#
+#     mip2 = mgmt_root.tm.sys.management_ips.management_ip.create(
+#         name='172.16.44.16/24')
+#     assert mip2.name == '172.16.44.16/24'
+
+
+def test_modify_addr(request, mgmt_root):
```
wojtek0806: this needs changing due to update/modify being possible

f5/bigip/tm/sys/test/functional/test_management_ip.py
```diff
@@ -0,0 +1,109 @@
...
+    name = '10.0.2.15/24'
+    mip1 = mgmt_root.tm.sys.management_ips.management_ip.load(
+        name=name)
+    mip1.name = '10.0.2.16/24'
+    with pytest.raises(UnsupportedMethod) as err:
+        mip1.modify()
+    assert 'Management_Ip does not support the modify method' in str(err)
+
+
+def test_update_addr(request, mgmt_root):
```
wojtek0806: this needs changing do to the update being possible

wojtek0806 (changes requested):
See my inline comments

wojtek0806 (approved):

wojtek0806 (changes requested):
see my inline comments and correct accordingly

f5/bigip/tm/sys/management_ip.py
```diff
@@ -0,0 +1,62 @@
...
+    """BIG-IP® management-ip resource"""
+    def __init__(self, Management_Ips):
+        super(Management_Ip, self).__init__(Management_Ips)
+        self._meta_data['required_json_kind'] = \
+            'tm:sys:management-ip:management-ipstate'
+        self._meta_data['required_creation_parameters'].update(('name',))
+
+    def load(self, **kwargs):
+        kwargs['transform_name'] = True
+        return self._load(**kwargs)
```
wojtek0806: nice!

jasonrahm:
ready for your re-review @wojtek0806

wojtek0806 (approved):
Looks good now, we can add 'create' test later, for Lack of 2 NAT interfaces inside VirtualBox in Vagrant is not helping. I need to talk to @caphrim007 about creating and using NAT network, this way you can have multiple interfaces with NAT on them

PR #1014, opened 2017-02-22
Fixes #959

### LINKED ISSUE #959: drop python 2.6 support from travis

wojtek0806:
Problem:
Python 2.6 on Travis is depracated

Analysis:
We drop testing for python 2.6 on Travis with this patch, also previous disabled properties method is enabled

Tests:
Unit
Flake8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ services:
   - docker
 python:
   - '2.7'
-  - '2.6'
 before_install:
   - git config --global user.email "OpenStack_TravisCI@f5.com"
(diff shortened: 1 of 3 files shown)


wojtek0806:
hmm this does not do what I would expect it to

wojtek0806:
bounced this to see if the test is skipped, as it should run, and python 2.6 testing should not be executed

wojtek0806:
i like to live on the edge, so I am merging this :) feel free to frown upon later (for disabling py 2.6)

PR #1040, opened 2017-03-10
Extract fixtures

zancas:
Folks, we're migrating "externally useful" test infrastructure (fixtures) into their own repository.

This will among other things:

(0) allow them to evolve semi-independently of the sdk
(1)  make their consumers explicitly ask for them

I think a large fraction of SDK developers use it for test, perhaps exclusively in some cases.  Promoting `pytest_f5sdk` to it's own project may cause those fixtures (that use case) to mature/evolve more quickly.
I'm excited about the prospects.
An effect that is particularly relevant here is that the test code _within_ the SDK (the stuff that remains) will likely be simpler.

So, you may be asking yourself....  "Why am I a reviewer of this commit?"

Good question!

The answer is:   Because I want to highlight the change in where we put test infrastructure code.

Y'all are the people most likely to generate new SDK tests.


caphrim007:
ansible did this same thing, albeit with their modules, and it was a bad idea because to make a change in one required that a totally separate PR be created for updating the other.

This led to PR problems because a change in one repo essentially broke the other repo until the dev could get around to writing another PR to un-break the repo that was affected.

Ansible solved it, in the end, by merging the separate repos back in to one repo. Dunno if I see any problems with the current model

caphrim007:
Also, you missed the fixtures in bigiq and iworkflow.

https://github.com/F5Networks/f5-common-python/blob/development/f5/iworkflow/conftest.py

we use the design of pytest to specifically override the mgmt_root fixture so that it is relevant to the product under test. Otherwise you would need to have more options like iworkflow_mgmt_root, bigiq_mgmt_root, etc

wojtek0806:
totally agree with @tar

zancas:
Good points @tar.

  Let me start by clarifying something.

  It's _NOT_ the intention to remove _ALL_ (or even most) fixtures from the SDK. 
  The goal of this change to decouple *generally useful* fixtures from the SDK itself.

To the point about fixture composition:
  Suppose you have a fixture that uses pytest-style fixture composition, to generate a "more custom" fixture appropriate to a specific test regime (e.g. a product).   If that custom fixture is not generally (externally) useful, that is, if there's not another project (git project) that can productively use that fixture...   then that custom fixture *never needs to change*.

Concretely, you have some code like this:

```Python
@pytest.fixture
def my_custom_mgmt_root(mgmt_root):
     """This custom mgmt_root is different from 'mgmt_root', because it's CUSTOM!"""
     BLAH BLAH
``` 

NOTHING should change here.    The fact that the `mgmt_root` fixture definition has moved won't require any modification of this code.
Does everyone see this?
This is important, I wouldn't try to drop a change that required that kind of refactor on someone else.

In regards to the specific example:
  Per the customization of the "IWorkFlow" `mgmt_root` fixture, I don't see that being affected.  In particular *THAT CODE DOESN'T NEED TO CHANGE* (because of this migration).
    CAVEAT:  I think there's a tweak to apply to that code. I opened an issue to discuss details here:
      https://github.com/F5Networks/f5-common-python/issues/1042

Now, per the interdependency issues:
   I think you're saying:
   Having 2 tightly coupled projects such that changing 1 requires a change in the other, is a bad thing.  I agree totally.   I'm not moving us in that direction (at least, not very far).

   So let's talk about coupling.  The `pytest_f5sdk` Python distribution won't be tightly coupled to the `f5-common-python` distribution.

   There are two directions to consider in this relationship:
   (0) `f5-common-python`-depends-on-`pytest_f5sdk`
   (1)  `pytest_f5sdk`-depends-on-`f5-common-python`

   In the `f5-common-python`-depends-on-`pytest_f5sdk` direction the only dependency is a test dependency.   We currently implement this by adding a line to `requirements.test.txt` that points to the `https://github.com/F5Networks/pytest_f5sdk.git`.   This is, in turn, "pip installed" by tox into the various test environments.  So, the _only_ dependency on the `pytest_f5sdk` is a _test_ dependency, and notice it points to `master` meaning....    as soon as a new fixture is added to `pytest_f5sdk` it's automagically available to `f5-common-python`'s test regime (assuming your tox environment isn't stale.   AND the SDK functionality itself, the core logic, has 0 dependency on `pytest_f5sdk`.

   Now, going the other direction, `pytest_f5sdk`-depends-on-`f5-common-python`, this will generally be managed by the distribution that `pytest_f5sdk` has been sistered with.   For example, suppose you're testing `f5-common-python`, in that case tox will install the development commit you're working on, and when `pytest_f5sdk` is evaluated for installation `setuptools` will see that the `f5-sdk` is installed and do nothing. 

  Semantic versioning will ensure version compatibility, and if we need to change code in the `pytest_f5sdk` to handle a backwards compatibility break from the `f5-sdk` that is work that would be required anyway (it's just done in a different location).


  So, there's *some* minimal overhead (cost) implied by this change...   what's the benefit?

  To the OpenStack Team the benefit is immediate and significant.

   A good example of this benefit is the logic in the device registration setup and teardown code.   We now have a single source of truth for this logic that can evolve with little constraint introduced by the version of the SDK proper.

   Pre-change, when we wanted to update the [AGENT_LB_DEL_ORDER](https://github.com/F5Networks/f5-common-python/blob/e8bdc136255aad7938ed65260ffe50cfe3553a39/f5/utils/testutils/registrytools.py#L25) dictionary that we use to cleanup after tests that create loadbalancers, we had to release a new version of the SDK and update the sister (f5-openstack-agent) to depend on the new version!!!   Obviously this is bad.    Now all we have to do is add a new commit to `pytest_f5sdk`, and rerun tox!

   OK now I'm going to just scratch a few things for later:

   Other benefits:

      f5-sdk get's the flexibility also
      more/different eyes on pytest_f5sdk

BTW, a lot of this architecture is due to @pjbreaux 's comments.

zancas:
Oh, also I should mention that I _AM_ flexible on this work.   We don't necessarily have to solve our coupling problem in this way....    BUT the OS Team does need to solve it...   so if we reject this solution it will have to be in favor of a better/competitive/alternate one.

wojtek0806:
@zancas @pjbreaux @caphrim007 

I will add something here which may/ or not be related. I think in our tests we should start using fixtures for resource setup/teardown. I have been playing with it in my ASM functional tests:

https://github.com/F5Networks/f5-common-python/blob/development/f5/bigip/tm/asm/test/functional/test_policies.py#L101

I have found that in the case of ASM policies this saves a ton of time and improves performance, less API calls too. 

So just so you know how am I using mgmt_root, maybe not the same way you guys thinking here but worth remembering this when you start moving stuff away from this SDK.

If you agree with this approach, we can discuss some principles (given that my approach was experimenting with scopes (for one reason or the other)), and create some sort of guidelines for developer guide. Then proceed to rewriting our current tests to reflect the agreed test framework/model.

caphrim007:
@wojtek0806 the bigiq and iworkflow modules already use such a process

wojtek0806:
@caphrim007  hah i did not know that, well great minds and all that ;)

wojtek0806:
@zancas @caphrim007 @pjbreaux 

I am wondering if maybe moving those kind of utilities as registrytools.py, would be useful, rather than removing of mgmt_root fixtures.  I base this reasoning on the fact that we don't really use this in SDK testing anyway. I can imagine why you guys use such tools (failing tests (due to code errors) that leave some objects on the LBs. However using fixtures like this:

https://github.com/F5Networks/f5-common-python/blob/development/f5/bigip/tm/security/test/functional/test_firewall.py#L59

,would ensure that no artifacts are left over.  Is there any particular reason why you haven't started to utilize test fixtures in such a way? Since I am not involved in OS I am not sure how complex would be.

So if we were to move in our testing suite in that direction, then **I could live with decoupling the most commonly used fixtures that would never/rarely change to another repository**, as well as some of the tools that we never/barely use in SDK testing. 

Personally I would like to rework our entire testing suite in here, to standardize it and have some established patterns, as for now this is all over the place. I love consistency as you have noticed so we need to discuss this. I will open another issue for that.

wojtek0806 (commented):
@zancas @caphrim007 @pjbreaux

I am wondering if maybe moving those kind of utilities as registrytools.py, would be useful, rather than removing of mgmt_root fixtures. I base this reasoning on the fact that we don't really use this in SDK testing anyway. I can imagine why you guys use such tools (failing tests (due to code errors) that leave some objects on the LBs. However using fixtures like this:

https://github.com/F5Networks/f5-common-python/blob/development/f5/bigip/tm/security/test/functional/test_firewall.py#L59

,would ensure that no artifacts are left over. Is there any particular reason why you haven't started to utilize test fixtures in such a way? Since I am not involved in OS I am not sure how complex would be.

So if we were to move in our testing suite in that direction, then I could live with decoupling the most commonly used fixtures that would never/rarely change to another repository, as well as some of the tools that we never/barely use in SDK testing.

Personally I would like to rework our entire testing suite in here, to standardize it and have some established patterns, as for now this is all over the place. I love consistency as you have noticed so we need to discuss this. I will open another issue for that.

caphrim007:
@zancas there's another f5'er around here that is using pytest here for his own project, so in the scope of there being now 3 (us, ansible, and him) that might use these fixtures, i would now be in favor of extracting the general fixtures to their own package.

as long as @jasonrahm @wojtek0806 and I are able to merge to that repo

PR #1043, opened 2017-03-12
Fixes #1021

### LINKED ISSUE #1021: replace set() declarations with set literals
given that we removed 2.6 from the menu, we can now start refactoring set declarations to use more efficient set literals. 

Instead of set(('item1', 'item2')) we will use {item1,item2} syntax

wojtek0806:
Problem:
We were using set(()) constructors in places where they could have been easily replaced by more effective set literals.

Analysis:
With removal of py2.6 support, we have lost the need to having the old set() syntax, this is now corrected to more effective set literals.

Tests:
Functional
Unit
Flake8

### DIFF
--- a/f5/bigip/resource.py
+++ b/f5/bigip/resource.py
@@ -827,9 +827,9 @@ class Resource(ResourceBase):
         super(Resource, self).__init__(container)
         self._meta_data.pop('uri')
         # Creation fails without these.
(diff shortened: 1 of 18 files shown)


wojtek0806:
Can you verify this did not break anything in your bigiq and iworkflow? and merge after review?

@caphrim007

Issue #840, opened 2016-12-06
Invalid state when updating a node


erjac77:
**What did you do?**

Updated the description of a node.

**What did you expect to see?**

Node successfully updated.

**What did you see instead? Under which circumstances?**

I got an invalid property value error message for the state attribute.

```
Text: u'{"code":400,"message":"invalid property value \\"state\\":\\"checking\\"","errorStack":[]}'
```

**Steps to reproduce the issue**

```
from f5.bigip import ManagementRoot
mgmt = ManagementRoot("x.x.x.x", "admin", "admin")

mynode = mgmt.tm.ltm.nodes.node.create(name='mynode', partition='Common', address='11.11.11.11', monitor="/Common/icmp")
mynode.description = "my node"
mynode.update()

/usr/local/lib/python2.7/dist-packages/requests/packages/urllib3/connectionpool.py:838: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/security.html
  InsecureRequestWarning)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/f5/bigip/tm/ltm/node.py", line 82, in update
    return self._update(**kwargs)
  File "/usr/local/lib/python2.7/dist-packages/f5/bigip/resource.py", line 511, in _update
    response = session.put(update_uri, json=data_dict, **requests_params)
  File "/usr/local/lib/python2.7/dist-packages/icontrol/session.py", line 241, in wrapper
    raise iControlUnexpectedHTTPError(error_message, response=response)
icontrol.exceptions.iControlUnexpectedHTTPError: 400 Unexpected Error: Bad Request for uri: https://x.x.x.x:443/mgmt/tm/ltm/node/~Common~mynode/
Text: u'{"code":400,"message":"invalid property value \\"state\\":\\"checking\\"","errorStack":[]}'
```

As per the documentation, the only acceptable values for state are:
`state [user-down | user-up]`

The SDK should check if the returned value is valid before sending the rest call. If not, the state attribute can be omitted.

wojtek0806:
@erjac77 

I think if you used modify() instead this would not happen, when we do update() we use PUT, which forces bigip to fill in the blanks for the rest of the values with defaults, which would probably lead to the issue. So what happens when you just use modify() ?

So that 'state:checking' was probably set by BIGIP internally, resulting in error.

[...]

PR #1064, opened 2017-03-17
Fixes #985

### LINKED ISSUE #985: Invalid state when updating a pool and pool member?
this needs investigating as per comment in #840

wojtek0806:
Problem:
Pool Members were suffering from the same issue with state/session setting, as described in #840

Analysis:
This PR brings the user experience of pool members in line with that of nodes. Minor adjustments made to node.py to add some defaults to prevent key errors raising during attribute pop

Tests:
Functional
Flake8


wojtek0806:
@tar please review and merge

PR #1111, opened 2017-04-10
Fixes #1056

### LINKED ISSUE #1056: required create parameter checking needs enchancement
Currently create parameter checking does not include scenarios where we have 1 of x mandatory parameters required, this leads to unnecessary duplication when overloading create() methods in the resources. 

The required method __minimum_one_is_missing is already in the resources.py, just _create() needs a bit of refactoring to include that check.

We would also need to add another _meta_data entry to keep a record of those parameters for each resource. By default this should be empty set().

wojtek0806:
Problem:
Streamlines additional mandatory params checking

Analysis:
Incorporated _missing_one_required method usage into the Resource class, as more and more endpoint required it. Tweaked some policy tests as they failed intermittently.

Tests:
Flake8
Unit
Functional

### DIFF
--- a/f5/bigip/resource.py
+++ b/f5/bigip/resource.py
@@ -146,28 +146,6 @@ def _missing_required_parameters(rqset, **kwargs):
         return list(required_minus_received)
 
 
-def _minimum_one_is_missing(rqset, **kwargs):
-    """Helper function to do operation on sets
-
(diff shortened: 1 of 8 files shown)


wojtek0806:
@jasonrahm please review and merge :)

PR #1114, opened 2017-04-11
The testing of non-python files was breaking tox

caphrim007:
Issues:
Fixes #<issueid>

Problem:
The testing of non-python files was breaking tox

Analysis:
non-python file's should be tested via their directory

Tests:


caphrim007:
@wojtek0806

PR #1138, opened 2017-05-23
Members modify() broken when setting session attribute

### LINKED ISSUE #1137: Members modify() broken when setting session attribute
The modify command incorrectly checks value of "state" attribute when "session" attribute is passed in modify(). See line 217 in pool.py:

`        if 'session' in patch:
            if patch['session'] != 'user-enabled' and patch['state'] != \
                    'user-disabled':`

patch['state'] should be patch['session']

This is causing an error in the F5 OpenStack agent when users try to set admin state up/down for a pool member.

jlongstaf:
Issues:
Fixes #1137

Problem: Member modify incorrectly checks value of 'state' attribute
when user passes 'session' attribute to modify.

Analysis: patch['state'] is not correct. should be patch['session']

Tests: Added new test, test_session_modify(), to match existing test
test_state_modify().

### DIFF
--- a/f5/bigip/tm/ltm/pool.py
+++ b/f5/bigip/tm/ltm/pool.py
@@ -214,7 +214,7 @@ class Members(Resource):
                       "'user-down'" % patch['state']
                 raise MemberStateModifyUnsupported(msg)
         if 'session' in patch:
-            if patch['session'] != 'user-enabled' and patch['state'] != \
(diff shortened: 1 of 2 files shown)


pjbreaux (approved):
Looks good.

caphrim007:
lgtm

PR #1150, opened 2017-07-17
Adds partition support to SDK.

### LINKED ISSUE #1149: add partition support
Folder support exists, but folders are not partitions

caphrim007:
Somebody added folder support, but folders are not partitions as
recognized by TMUI. So this adds proper partition API support

Issues:
Fixes #1149

Problem:
Partition support did not exist in the SDK

Analysis:
This patch adds it

Tests:
* functional
* unit


jasonrahm:
ok, this looks good to me.

PR #1164, opened 2017-08-10
Adds support for the QUERY operation

### LINKED ISSUE #1163: iapp endpoint requires a package file even when it shouldnt
the required creation params for the iapp endpoint are incorrect and need to support QUERY operation

caphrim007:
Issues:
Fixes #1163

Problem:
Previously, the required creation params were too strict and did
not allow for QUERY operations to be created

Analysis:
This patch fixes that while at the same time retaining the previous
requirements for INSTALL operations

Tests:
functional
unit


jasonrahm (approved):
lgtm

PR #1195, opened 2017-08-22
Makes the rules attribute return an empty list if it is not set

### LINKED ISSUE #1170: Return an empty list rather than AttributeError (class 'f5.bigip.tm.ltm.virtual.Virtual.rules')
Hello pyGurus!

Originating from answer at DevCentral (https://devcentral.f5.com/questions/need-help-on-f5-automation-to-read-the-irules-attached-t-all-the-virtual-servers-55140). Accepted answer contains a code sample.

When Virtual Server has no iRules attached to it, an empty list should be returned when 'vserver.rules' is called. Today it returns `AttributeError: '<class 'f5.bigip.tm.ltm.virtual.Virtual'>' object has no attribute 'rules'`.

This is problematic in for-cycle iterations. If one Virtual Server happens to have no iRules, then script work is prematurely terminated

Regards,

caphrim007:
Issues:
Fixes #1170

Problem:
If there are no rules for a Virtual, the API will return no attribute.
This causes problems when wanting to loop through that.

Analysis:
So instead, the SDK will return just an empty list so that you dont need to catch
AttributeError exceptions

Tests:
functional

### DIFF
--- a/f5/bigip/tm/ltm/test/functional/test_virtual.py
+++ b/f5/bigip/tm/ltm/test/functional/test_virtual.py
@@ -24,18 +24,23 @@ TESTDESCRIPTION = "TESTDESCRIPTION"
 
 
 @pytest.fixture
-def virtual_setup(request, mgmt_root):
+def basic_rule(mgmt_root):
+    rule1 = mgmt_root.tm.ltm.rules.rule.create(
+        name='foo',
+        partition='Common',
+        apiAnonymous=''
+    )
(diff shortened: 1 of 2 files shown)


caphrim007:
this is a breaking change so will need a rev of the major version of the SDK

wojtek0806:
as discussed this might break backwards compatibility, so it will be a warrant for a major release, merging

Issue #1252, opened 2017-08-24
Add load as an allowable command for sys/config


jasonrahm:
save is the only allowable command for sys/config but load should be supported as well.

caphrim007:
fixed

PR #1253, opened 2017-08-24
Issue #1252: Add load as allowable command for sys/config

jasonrahm:
Single line change, appending load as allowed_commands metadata


caphrim007:
lgtm

PR #1258, opened 2017-08-28
Updates requirements and test settings

### LINKED ISSUE #1257: Change tests and requirements
To enable better python 2 and 3 testing. Changes to requirements and tox file need to be made

caphrim007:
Issues:
Fixes #1257

Problem:
Requirements and tox updates needed to be made for bugfixes in
software and to disable python caches for testing

Analysis:
This patch does the above

Tests:

### DIFF
--- a/requirements.test.txt
+++ b/requirements.test.txt
@@ -3,7 +3,7 @@
 # Test Requirements
 hacking==0.13.0
 mock==2.0.0
-pytest==3.0.5
+pytest==3.2.1
 pytest-cov>=2.2.1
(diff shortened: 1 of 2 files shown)


wojtek0806:
all fine, i am wondering why 3.2 not the latest 3.x is that larger project req. ? never paid attention to this before frankly so i am curious :)

caphrim007:
@wojtek0806 3.2.1 is the latest, and we pinned to a release in the past, so i continued to pin to a release when i updated it.

Issue #1152, opened 2017-07-21
asm export policy task should allow inline exports


wojtek0806:
at the moment we only force the exports to a file, asm policy can be exported in an inline format

wojtek0806:
this will need to be implemented for ansible asm_policy_fetch module

wojtek0806:
this is merged

PR #1271, opened 2017-09-06
Fixes #1153, #1152

### LINKED ISSUE #1153: asm policy import should allow policyTemplateReference property
this needs implemented for ansible asm_policy_import module

wojtek0806:
Problem:
Importing policy from policy templates was not supported. Exporting policy as inline was not supported

Analysis:
Added importing from policy templates and exporting to inline

Tests:
Flake8
Unit
Functional


jasonrahm (approved):
lgtm

PR #1293, opened 2017-09-18
Adds documentation for (manual) release process

### LINKED ISSUE #1292: add documentation for manual release process
because it doesnt exist and nobody understands the automated process that is being done.

caphrim007:
Issues:
Fixes #1292

Problem:
The release process was not documented and if the automatic process
fails, there is no insight into how it worked in the first place,
and therefore how to fix it.

Analysis:
This patch includes docs that explain the manual process that can
be used if the automatic process fails or is not understood.

Tests:

### DIFF
--- a/docs/devguide.rst
+++ b/docs/devguide.rst
@@ -1,4 +1,288 @@
 Developer Guide
 ===============
 
-**COMING SOON**
+
+
+Releasing new versions
+----------------------
+
+Releasing new versions can be done in two different ways
+
+* automatically
+* manually
+
+In this document I'll outline the steps required for manual release.
+
+Version commit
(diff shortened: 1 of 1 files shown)


docs/devguide.rst
```diff
@@ -1,4 +1,288 @@
...
+.. code-block:: bash
+
+   git tag v3.0.2
+
+And also push that tag upstream because it is what you will tell Github to release
+from.
+
+.. code-block:: bash
+
+   git push v3.0.2 upstream
```
zancas, line 88: Strictly speaking this is unnecessary since the release notes interface adds a tag.

zancas:
LGTM

PR #1299, opened 2017-09-28
Fixes session errors

### LINKED ISSUE #1298: node session data works incorrectly if state is missing
The node code is doing a session + state check when it should really be doing a session + session check. Additionally, there are missing session values and python3 errors.

caphrim007:
Issues:
Fixes #1298

Problem:
Session values were being compared, incorrectly, to state values.
Python3 errors
Missing state values

Analysis:
This fixes the above errors

Tests:
* functional


jasonrahm (approved):
lgtm

Issue #1063, opened 2017-03-17
sanity checks on update()


wojtek0806:
Currently most of our sanity checks for parameters rely on **kwargs passed to update(), those sanity checks do not work when the user adds object attributes directly:

````
r1.description == 'SOMETHING'
r1.update()

vs

r1.update(description='SOMETHING')

```

PR #1308, opened 2017-10-11
Handle rule descriptions according to BIG-IP version

### LINKED ISSUE #1307: Error when setting policy rule "description" on BIG-IP v11.6.1
Setting the "description" parameter on a policy rule succeeds on BIG-IP v12.1.0 but fails on v11.6.1:

"code":400,"message":"Found unexpected json pair at configuration item /ltm/policy/~cf~cf-routing-policy/rules/cf-uaa-724b8d651b68119b. The json pair is \\"description\\":\\"route: uaa.sys.10.190.24.189.xip.io\\".","errorStack":[]

The SDK isn't detecting that this parameter is not valid for v11.6.1, but it should.

edarzins:
Issues:
Fixes #1307

Problem:
The 'description' parameter does not exist in policy rules prior
to BIG-IP v12.1.0. Attempting to set the rule description for these
earlier BIG-IP versions results in an error response.

Analysis:
If the BIG-IP version is earlier than v12.1.0, strip the description from
policy rule.

Tests:
test_rules_description (f5/bigip/tm/ltm/test/functional/test_policy.py)

### DIFF
--- a/f5/bigip/tm/ltm/policy.py
+++ b/f5/bigip/tm/ltm/policy.py
@@ -70,6 +70,8 @@ class Policy(Resource):
         legacy = kwargs.pop('legacy', False)
         publish = kwargs.pop('publish', False)
         if LooseVersion(tmos_ver) < LooseVersion('12.1.0'):
+            for r in kwargs.get('rules', []):
+                r.pop('description', None)
             return super(Policy, self)._create(**kwargs)
         else:
             if legacy:
@@ -94,6 +96,9 @@ class Policy(Resource):
 
         legacy = patch.pop('legacy', False)
         tmos_ver = self._meta_data['bigip']._meta_data['tmos_version']
+        if LooseVersion(tmos_ver) < LooseVersion('12.1.0'):
+            for r in patch.get('rules', []):
+                r.pop('description', None)
         if 'Drafts' not in self._meta_data['uri'] and \
                 LooseVersion(tmos_ver) >= LooseVersion('12.1.0') and \
                 not legacy:
@@ -110,6 +115,9 @@ class Policy(Resource):
 
         legacy = kwargs.pop('legacy', False)
         tmos_ver = self._meta_data['bigip']._meta_data['tmos_version']
+        if LooseVersion(tmos_ver) < LooseVersion('12.1.0'):
+            for r in kwargs.get('rules', []):
+                r.pop('description', None)
         if 'Drafts' not in self._meta_data['uri'] and \
                 LooseVersion(tmos_ver) >= LooseVersion('12.1.0') and \
                 not legacy:
--- a/f5/bigip/tm/ltm/policy.py
+++ b/f5/bigip/tm/ltm/policy.py
@@ -36,6 +36,9 @@ from f5.sdk_exception import NonExtantPolicyRule
 from f5.sdk_exception import OperationNotSupportedOnPublishedPolicy
 
 from distutils.version import LooseVersion
+import logging
+
+logger = logging.getLogger(__name__)
 
 
 class Policys(Collection):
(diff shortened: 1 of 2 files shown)


f5/bigip/tm/ltm/policy.py
```diff
@@ -94,6 +96,9 @@ def _modify(self, **patch):
 
         legacy = patch.pop('legacy', False)
         tmos_ver = self._meta_data['bigip']._meta_data['tmos_version']
+        if LooseVersion(tmos_ver) < LooseVersion('12.1.0'):
+            for r in patch.get('rules', []):
+                r.pop('description', None)
```
wojtek0806: If we are going to make this change, and what it seems this might be one of many such "exceptions", it would be prudent to have this folded into _meta_data, for example _meta_data['optional_parameters'], similarly to:

https://github.com/F5Networks/f5-common-python/blob/development/f5/bigip/resource.py#L839

Also if we silently remove the description from **kwargs how would the caller know something went wrong? If the caller does not know that a certain attribute does not exist in the specific version, we should inform him by raising exception with appropriate method, rather than just removing.

The only exception to such rule is here, and this is due to the fact that we send the entire __dict__ (minus __meta_data) to the device when we do update(), and state/session parameters are read only in specific situations, so the caller need not to be notified, unless the caller attempt to set them to an illegal value:

https://github.com/F5Networks/f5-common-python/blob/development/f5/bigip/tm/ltm/node.py#L95
https://github.com/F5Networks/f5-common-python/blob/development/f5/bigip/tm/ltm/pool.py#L131

We raise if the parameter is set to illegal value by the caller:
https://github.com/F5Networks/f5-common-python/blob/development/f5/bigip/tm/ltm/pool.py#L209

wojtek0806 (changes requested):
Please rework this as per my comments

f5/bigip/tm/ltm/policy.py
wojtek0806: The above behavior of update() will be modified once i get a chance, the issue is already opened up for it:
https://github.com/F5Networks/f5-common-python/issues/1063

f5/bigip/tm/ltm/policy.py
```diff
@@ -56,6 +59,24 @@ def __init__(self, policy_s):
...
+        filter these out for earlier versions to allow backward comatibility.
+        '''
+
+        if LooseVersion(tmos_ver) < LooseVersion('12.1.0'):
+            for k, parms in self._meta_data['optional_parameters'].items():
+                for r in kwargs.get(k, []):
+                    for parm in parms:
+                        value = r.pop(parm, None)
+                        if value is not None:
+                            logger.info(
```
wojtek0806, line 77: I would suggest raising an error, it is easier to handle by the caller

f5/bigip/tm/ltm/policy.py
```diff
@@ -56,6 +59,24 @@ def __init__(self, policy_s):
         self._meta_data['required_creation_parameters'].update(('strategy',))
         temp = {'tm:ltm:policy:rules:rulescollectionstate': Rules_s}
         self._meta_data['attribute_registry'] = temp
+        self._meta_data['optional_parameters'] = {'rules': ['description']}
+
+    def _filter_version_specific_options(self, tmos_ver, **kwargs):
```
wojtek0806, line 64: i think the point was missed here, i was talking about streamlining this into resource.py, but fine since this is a bigger job, just open an issue for it and assign it to me, i will do this later with update() rework

wojtek0806 (changes requested):
i would remove logging and raise exception, after that is done this can be merged

f5/bigip/tm/ltm/policy.py
edarzins: I discussed this with @pjbreaux and @zancas and we agreed that the caller should not receive an exception. SDK should handle the BIG-IP version compatibility issue by stripping the optional parameter and logging the event.

wojtek0806: This is taken from our dev playbook, which is what we are adhering to, so what has changed? specifying a non existent parameter is a failure

````
Failures generate exceptions
    If the SDK enters a surprising or unexpected state it raises an exception. That's it. It's not generally up to the SDK to implement decision logic that handles edge-cases.. EXCEPT where the SDK is smoothing known issues in the device REST server. (See below.)

````

caphrim007:
I look at the description issue as a form of deprecation. Deprecation needs raise no exception. Indeed, the code will continue on, but along the way will need to tweak things because of the deprecated code path

So for now I'm ok with this

PR #1316, opened 2017-11-01
Adds the ability for ASM tasks to import to different partititions

### LINKED ISSUE #1315: ASM task import needs to support partitions
currently you cannot import an ASM policy to a different partition

caphrim007:
Issues:
Fixes #1315

Problem:
Previously you could not import an ASM policy to a specific partition

Analysis:
This patch provides two methods to import an ASM policy to a partition.

First, you can specify the fullPath argument. This is directly mapped
to the fullPath attribute of the resource and ASM will import the partition
that is specified in the full path. ex, /Partition1/my_policy

Second, you can specify a combination of 'name' and 'partition' to the
`create` method. This method is just a convenience wrapper on top of
the aforementioned method. In the end, this method will generate the
fullPath for you. This method is more inline with other APIs that
accept a 'partition' argument

Tests:
functional

### DIFF
--- a/f5/bigip/tm/asm/tasks.py
+++ b/f5/bigip/tm/asm/tasks.py
@@ -30,7 +30,10 @@ from f5.bigip.resource import AsmResource
 from f5.bigip.resource import AsmTaskResource
 from f5.bigip.resource import Collection
 from f5.bigip.resource import OrganizingCollection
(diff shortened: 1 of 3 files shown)


jasonrahm (approved):
lgtm, @caphrim007

PR #1326, opened 2017-11-22
Adds auth_provider

### LINKED ISSUE #1325: support auth_providers for bigiq and bigip
the `token` abstraction that we have is not quite the right abstraction. It should instead be auth_provider. BIG-IQ provides a facility for this. BIG-IP does too, it's called "tmos", but we incorrectly just say `token=true`.

Add support for auth_provider while maintaining legacy support for token

caphrim007:
Issues:
Fixes #1325

Problem:
auth_provider specification was not available. Therefore, it was not possible
to authenticate to a BIG-IQ using different remote auths

Analysis:
This patch adds that after adding a patch to the f5-icontrol-rest-python repo

Tests:


caphrim007:

caphrim007:

caphrim007:

PR #1331, opened 2017-11-30
Adds authn/root API endpoint

### LINKED ISSUE #1330: add support for authn API
you can reset the root password with this api

echo "Reset root password back to default"
curl -k -u admin:admin -X POST -H "Content-Type: application/json" \
  https://localhost:10443/mgmt/shared/authn/root \
  -d '{"oldPassword": "f5site02", "newPassword": "default"}'

caphrim007:
Issues:
Fixes #1330

Problem:
The authn/root API was not available. This API is really weird because its
more like a fire-and-forget API. It provides no kind or selfLink in its
return value.

Analysis:
Had to work around the resource deficiencies. Introduced our own kind
and selfLink values because BIG-IP does not provide them. All methods
except create are invalid.

Tests:
functional
unit

### DIFF
--- a/f5/bigip/shared/__init__.py
+++ b/f5/bigip/shared/__init__.py
@@ -16,6 +16,7 @@
 #
 
 from f5.bigip.resource import OrganizingCollection
+from f5.bigip.shared.authn import Authn
 from f5.bigip.shared.authz import Authz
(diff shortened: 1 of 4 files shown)


caphrim007:

caphrim007:

caphrim007:

PR #1333, opened 2017-11-30
Adds hard timeout for management initialization

### LINKED ISSUE #541: Management initialization should allow for connection timeout setting
The ManagementRoot object allows you to pass a timeout=n kwarg, but it does not pass that kwarg to the get_tmos_version() nor the underlying session.get() request. That leads to situations where only default timeouts are possible with MangaementRoot which won't work.

I added a fix in the feature.bigiq branch.

caphrim007:
Issues:
Fixes #541

Problem:
The management root initialization may hang for longer than the
specified timeout value

Analysis:
The requests library's underlying urllib3 code is doing retries when
it fails to connect to a host or the connection is taking a long time.
If you have a timeout set (for example 5), the default retries is 10
and therefore your timeout is actually 50 seconds.

We can technically change this by providing a custom Retry() object
to the `from requests.adapters import HTTPAdapter` adapter, but that
would cause this retry to affect __all__ the API calls. We really
only want to override the initial management root `_get_tmos_version`
call.

So we use the eventlet package to override this. If eventlet is not
installed, the package will not be able to enforce this hard timeout
and will pass on trying.

By default, eventlet is installed with f5-sdk

Tests:
functional

### DIFF
--- a/devtools/bin/create-test-list.py
+++ b/devtools/bin/create-test-list.py
@@ -59,7 +59,8 @@ def determine_files_to_test(product, commit):
     results = []
     build_all = [
         'setup.py', 'f5/bigip/contexts.py', 'f5/bigip/mixins.py',
-        'f5/bigip/resource.py', 'f5sdk_plugins/fixtures.py'
+        'f5/bigip/resource.py', 'f5sdk_plugins/fixtures.py',
+        'f5/bigip/__init__.py'
     ]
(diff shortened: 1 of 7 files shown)


caphrim007:

caphrim007:

caphrim007:

caphrim007:

PR #1341, opened 2017-12-13
Removes eventlet in favor of signal package

### LINKED ISSUE #1340: replace eventlet with signal
signal is a package that ships with python. i think the eventlet package may have a number of problems with it and we probably want to switch to signal as we can make use of its facilities just as well

caphrim007:
Issues:
Fixes #1340

Problem:
The eventlet package was a good first attempt, but seems to have
other issues that plague third party tools.

Analysis:
signal comes with Python and should not have the same issues

Tests:
functional

### DIFF
--- a/f5/bigip/__init__.py
+++ b/f5/bigip/__init__.py
@@ -24,16 +24,10 @@ except ImportError:
     import urllib.parse as urlparse
 
 try:
-    import eventlet
-
-    # Workaround for bug 401 addressed here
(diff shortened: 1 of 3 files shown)


caphrim007:
wait on merging this

caphrim007:

caphrim007:

caphrim007:

caphrim007:

Issue #1323, opened 2017-11-15
Unable to use the "exists" method for GTM Topology


erjac77:
Hi,

The exists method for GTM Topology records doesn't seem to work with a Datacenter as LDNS routing destination.

**_With the partition_** in the Datacenter's name, I get a "**_instance names and partitions cannot contain '/'_**" error message.
```
>>> topo = mgmt.tm.gtm.topology_s.topology.exists(name='ldns: country US server: datacenter /Common/DC1', partition='Common')       
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/f5/bigip/resource.py", line 1108, in exists
    return self._exists(**kwargs)
  File "/usr/local/lib/python2.7/dist-packages/f5/bigip/resource.py", line 1120, in _exists
    session.get(base_uri, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/icontrol/session.py", line 252, in wrapper
    **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/icontrol/session.py", line 200, in generate_bigip_uri
    **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/icontrol/session.py", line 168, in _validate_uri_parts
    _validate_name_partition_subpath(name)
  File "/usr/local/lib/python2.7/dist-packages/icontrol/session.py", line 138, in _validate_name_partition_subpath
    raise InvalidInstanceNameOrFolder(error_message)
icontrol.exceptions.InvalidInstanceNameOrFolder: instance names and partitions cannot contain '/', but it's: ldns: country US server: datacenter /Common/DC1
```

**_Without the partition_** in the Datacenter's name, I get a "**_Topologies must specify both regions: ldns: server:_**" error message.
```
>>> topo = mgmt.tm.gtm.topology_s.topology.exists(name='ldns: country US server: datacenter DC1', partition='Common')        
/usr/local/lib/python2.7/dist-packages/urllib3/connectionpool.py:858: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
  InsecureRequestWarning)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/f5/bigip/resource.py", line 1108, in exists
    return self._exists(**kwargs)
[...]

PR #1347, opened 2017-12-30
Fixes Issue #1323 - exists method failing on GTM topology records

jasonrahm:
Problem: The / and ~ characters are disallowed in sdk URLs

Resolution: Overload the exists method for topology by applying:
  kwargs['transform_name'] = True

### DIFF
--- a/f5/bigip/tm/gtm/topology.py
+++ b/f5/bigip/tm/gtm/topology.py
@@ -146,6 +146,10 @@ class Topology(Resource):
 
         return self._load(**kwargs)
 
+    def exists(self, **kwargs):
+        kwargs['transform_name'] = True
+        return self._exists(**kwargs)
+
     def refresh(self, **kwargs):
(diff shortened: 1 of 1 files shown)


caphrim007:

jasonrahm:
hold off on merge until I hear back from the bug reporter that the fix works for him. I confirmed the behavior and the fix locally, but want to make sure.

jasonrahm:
@caphrim007 this fixes the problem, but also allows the partition kwarg but likely should not for topology records. Should I pop that kwarg if applied to guard against unintentional issues?

caphrim007:
@jasonrahm yes. If this changes in the future we can address it with a version check

jasonrahm:
ok, I'll add that and push later. Thanks

caphrim007:

jasonrahm:
@caphrim007 ready for review

caphrim007:

PR #1367, opened 2018-01-19
Updating API docs for three endpoints

jasonrahm:
Adding API docs for these endpoints:

bigip/tm/sys/smtp_server
bigip/tm/security/analytics
bigip/tm/security/protocol_inspection


caphrim007:

PR #1368, opened 2018-01-19
More fixes for devices in DG

caphrim007:
Problem:
Supporting exists was not done. Still allowed for update and patch to
be called.

Analysis:
Added support for the exists method. Removed update and patch because
there is nothing reasonable to patch on a device

Tests:


caphrim007:

PR #1374, opened 2018-01-25
Fixing exclusive attributes bug in the Virtual Server resource

jasonrahm:
Notified of this issue on an article on DC. Simple fix in this file:

f5/bigip/tm/ltm/virtual_server.py

### DIFF
--- a/f5/bigip/test/unit/test_resource.py
+++ b/f5/bigip/test/unit/test_resource.py
@@ -291,16 +291,15 @@ class TestResourceCreate(object):
         assert 'disabled' not in kwargs['json']
 
     def test_reduce_boolean_same_value(self, fake_vs):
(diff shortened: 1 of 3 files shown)


jasonrahm:
@caphrim007 given we have a solution for boolean pairs, and it's been implemented on the endpoint this resource unit test was pointing to, is the unit test that's failing even useful here?

jasonrahm:
this is ready for review, @caphrim007

caphrim007:
Ansible modules need to be verified

jasonrahm:
Probably should account for vlansEnabled / vlansDisabled as well for exclusive attributes, that wasn't asked about but is present on that object.

caphrim007:
verified that the Ansible modules run without issue

caphrim007:
ansible still works

PR #1395, opened 2018-02-16
Turns off debug. Adds an option to enable it.

### LINKED ISSUE #1394: debug output overwhelms memory on long-running sdk usage
The debug output gathering should be turned off by default, and an option to enable it be provided.

caphrim007:
Issues:
Fixes #1394

Problem:
Debugging could overwhelm memory on long-running sdk usage.

Analysis:
this is a problem, so by default we disable debugging and provide
and option to enable it.

Tests:


jasonrahm (approved):
lgtm

PR #1411, opened 2018-03-20
Add support for auth_provider to iWorkflow ManagementRoot

buzz-tee:
Small change to enable auth_provider (e.g. LDAP/AD authentication for iWorkflow + REST proxy).
Cloning behaviour (if auth_provider else token) from f5.bigip.BaseManagement.

Tested with iWorkflow 2.3.0 (as REST proxy) and BIG-IP 13.0 (being proxied to).


caphrim007:
lgtm

PR #1421, opened 2018-04-11
Added f5.bigip.resource.Stats to the list of allowed_lazy_attributes …

buzz-tee:
…on f5.big.tm.ltm collection objects Nodes, Pools, Rules, Virtuals and Virtual_Address_s

This is to allow obtaining all item stats with a single request which may come in handy when stats on large collections are required and calling item.stats.load() in a loop causes massive REST calls and huge delay.
Note that entries have to be mapped by the application, suggested criteria are item.fullPath and stat_entry['nestedStats']['entries']['tmName']['description'].

PR #1422, opened 2018-04-12
Added the user APIs to f5/bigip/shared/authz.py, fixed issue in f5/bigip/cm/system.py

nberezin:
1. Added the user APIs to f5/bigip/shared/authz.py:
These APIs are very handy when, let's say, you want to change the password of the existing user by providing the password itself without calculating the encryptedPassword like you should do it with tm.auth.users.user:
```python
collection = mgmt_root.tm.auth.users.user.load(name='admin')
collection.update(encryptedPassword='$6$fAj6uBdO$FwyVZvxAQo6ZXrFz5keM/lTTFssdR1Y8jx.sRfwWbezGQa3AilBgwYD9JX0OLOrzrMz0nQrVj1Efv6LMAh4Ua0')
```

Now you can do:
```python
collection = mgmt_root.shared.authz.users_s
resource = collection.user.load(name='admin')
resource.modify(name='admin', password='f5site02')
```

2. Fixed UnregisteredKind issue in f5/bigip/cm/system.py with bigip v13.1.0 and above:
Right now while running `pytest f5/bigip/shared/test/functional/test_authz.py` on 13.1.0 the exception is raised:
```
f5.sdk_exception.UnregisteredKind: 'cm:system:authn:providers:tmos:authproviderstate' is not registered!

f5/bigip/resource.py:801: UnregisteredKind
```

### DIFF
--- a/f5/bigip/cm/system.py
+++ b/f5/bigip/cm/system.py
@@ -49,18 +49,33 @@ class Providers(OrganizingCollection):
 class Tmos_s(Collection):
     def __init__(self, providers):
         super(Tmos_s, self).__init__(providers)
-        self._meta_data['required_json_kind'] = 'cm:system:authn:providers:tmos:mcpremoteprovidercollectionstate'
+        if (self._meta_data['bigip']._meta_data['tmos_version'] < '13.1.0'):
+            # Starting from bigip v13.1.0 tmos resources 'kind' was changed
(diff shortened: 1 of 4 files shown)


nberezin:
Same thing as in other PR #1421 , build errors are not related to the PR files.

PR #1429, opened 2018-04-17
Added f5.bigip.resource.Stats to the list of allowed_lazy_attributes …

buzz-tee:
…on f5.big.tm.ltm collection objects Nodes, Pools, Rules, Virtuals and Virtual_Address_s

This is to allow obtaining all item stats with a single request which may come in handy when stats on large collections are required and calling item.stats.load() in a loop causes massive REST calls and huge delay.
Note that entries have to be mapped by the application, suggested criteria are item.fullPath and stat_entry['nestedStats']['entries']['tmName']['description'].

### DIFF
--- a/f5/bigip/tm/ltm/node.py
+++ b/f5/bigip/tm/ltm/node.py
@@ -29,6 +29,7 @@ REST Kind
 
 from f5.bigip.resource import Collection
 from f5.bigip.resource import Resource
+from f5.bigip.resource import Stats
 from f5.sdk_exception import NodeStateModifyUnsupported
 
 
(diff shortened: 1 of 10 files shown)


jasonrahm:
Hi @BastiG, is there a reason only these five endpoints are updated to allow for loading stats in a single call?

I think this is a good update @caphrim007, but should we consider a larger swath of objects or call it good for now and add as requested? Seems if we make a helpful change like this it should be global on the objects that support it.

buzz-tee:
Hi @jasonrahm , these five endpoints were populated in my lab system and produced meaningful output at first glance so I decided to start with those.
I agree, there are probably many more (and I didn't have the time e.g. to dig into profiles Org Coll). Time permitting I can revisit some of those. As my lab only has LTM deployed I won't be able to look at other modules.

jasonrahm:
ok, let me discuss and I'll get back to you with a plan.

jasonrahm:
ok @BastiG, looks good, merging away!

Issue #1445, opened 2018-05-22
Add security/nat endpoints (Firewall NAT)


bobyaaaaw:
BigIP 12.1.0 added support for AFM Firewall NAT features.

related endpoints needed are:
 security/nat/source-translation
 security/nat/destination-translation
 security/nat/policy

bobyaaaaw:
Please assign this to me. I've got this started, pull request coming soon.

abhishekthukaram:
Hi Bob
This has already been written and submitted , in local repository of gitswarm

PR #1451, opened 2018-05-30
Adding files for FW-NAT

abhishekthukaram:


f5/bigip/tm/security/nat.py
```diff
@@ -0,0 +1,174 @@
...
+            {'tm:security:nat:source-translation:source-translationstate':
+                Source_Translation}
+
+
+class Source_Translation(Resource):
+    def __init__(self, source_translations):
+        super(Source_Translation, self).__init__(source_translations)
+        self._meta_data['required_json_kind'] = \
+            'tm:security:nat:source-translation:source-translationstate'
+        self._meta_data['required_creation_parameters'].update(
```
wojtek0806, line 118: 'name' parameter is not required its inherited from Resource, please remove

f5/bigip/tm/security/nat.py
```diff
@@ -0,0 +1,174 @@
...
+            ('partition', 'name'))
+        self._meta_data['required_load_parameters'].update(
```
wojtek0806: same as above, 'name' is not required, remove it please

f5/bigip/tm/security/nat.py
```diff
@@ -0,0 +1,174 @@
...
+        return super(Rule, self)._modify(**kwargs)
+
+    def load(self, **kwargs):
+        return super(Rule, self)._load(**kwargs)
+
+    def delete(self, **kwargs):
+        return super(Rule, self)._delete(**kwargs)
+
+    def exists(self, **kwargs):
+        """Some objects when deleted still return when called by their
```
wojtek0806: I just checked and this collection does not exist in 11.6.3, so the docstring is incorrect.

f5/bigip/tm/security/nat.py
```diff
@@ -0,0 +1,174 @@
...
+            {'tm:security:nat:destination-translation:destination-translationstate':
+                Destination_Translation}
+
+
+class Destination_Translation(Resource):
+    def __init__(self, destination_translations):
+        super(Destination_Translation, self).__init__(destination_translations)
+        self._meta_data['required_json_kind'] = \
+            'tm:security:nat:destination-translation:destination-translationstate'
+        self._meta_data['required_creation_parameters'].update(
```
wojtek0806, line 140: remove 'name' please

f5/bigip/tm/security/nat.py
```diff
@@ -0,0 +1,174 @@
...
+            ('partition', 'name'))
+        self._meta_data['required_load_parameters'].update(
```
wojtek0806: remove 'name' please

wojtek0806 (changes requested):
Those collections seems to not exist pre 12.0, you need to ensure that you set the minimum version appropriately, in your case i believe this is 12.1.0:

https://github.com/F5Networks/f5-common-python/blob/development/f5/bigip/resource.py#L169

You set this attribute on a collection, for example:

https://github.com/F5Networks/f5-common-python/blob/development/f5/bigip/tm/asm/policies/vulnerabilities.py#L27

wojtek0806:
One last thing please squash commits after making the requested changes

f5/bigip/tm/security/nat.py
```diff
@@ -0,0 +1,171 @@
...
+    def __init__(self, rules_s):
+        super(Rule, self).__init__(rules_s)
+        self._meta_data['required_json_kind'] = \
+            'tm:security:nat:policy:rules:rulesstate'
+
+        self._meta_data['exclusive_attributes'].append(
+            ('place-after', 'place-before'))
+        self._meta_data['minimum_additional_parameters'] = {'place-before',
+                                                            'place-after'}
+        self.tmos_ver = self._meta_data['bigip'].tmos_version
```
wojtek0806: why do you define this here ?

f5/bigip/tm/security/nat.py
```diff
@@ -0,0 +1,171 @@
...
+        return super(Rule, self)._load(**kwargs)
+
+
+class Source_Translations(Collection):
+    """BIG-IP® AFM® Nat Source Translation collection"""
+
+    def __init__(self, nat):
+        super(Source_Translations, self).__init__(nat)
+        self._meta_data['allowed_lazy_attributes'] = [Source_Translation]
+        self._meta_data['minimum_version'] = '12.1.0'
```
wojtek0806: From what I see NAT endpoint in its entirety is not in before 12.1.x, remove this please, we will define the version in Nat OC

f5/bigip/tm/security/nat.py
```diff
@@ -0,0 +1,171 @@
...
+from f5.bigip.resource import OrganizingCollection
+from f5.bigip.resource import Resource
+
+
+class Nat(OrganizingCollection):
+    """BIG-IP® AFM® Nat organizing collection."""
+
+    def __init__(self, security):
+        super(Nat, self).__init__(security)
+        self._meta_data['allowed_lazy_attributes'] = [
```
wojtek0806: Define minimum version here

f5/bigip/tm/security/nat.py
```diff
@@ -0,0 +1,171 @@
...
+            ('partition',))
+
+
+class Destination_Translations(Collection):
+    """BIG-IP® AFM® Nat Destination Translation collection"""
+
+    def __init__(self, nat):
+        super(Destination_Translations, self).__init__(nat)
+        self._meta_data['allowed_lazy_attributes'] = [Destination_Translation]
+        self._meta_data['minimum_version'] = '12.1.0'
```
wojtek0806: remove this please

wojtek0806 (changes requested):
Please follow with the changes, hopefully this will be the last of it :)

bobyaaaaw:
Fixes #1445

f5/bigip/tm/security/nat.py
```diff
@@ -0,0 +1,169 @@
...
+        """We need to implement the custom exclusive parameter check."""
+        self._check_exclusive_parameters(**kwargs)
+        return super(Rule, self)._update(**kwargs)
+
+    def modify(self, **kwargs):
+        """We need to implement the custom exclusive parameter check."""
+        self._check_exclusive_parameters(**kwargs)
+        return super(Rule, self)._modify(**kwargs)
+
+    def load(self, **kwargs):
```
bobyaaaaw, line 92: There's no need to override the load, delete and exists methods here if your just passing args to the superclass.

wojtek0806: only create has the exclusive parameter check implemented, so this is correct, you need to overload this to include the exclusive parameter call. I guess if this is repeated in many endpoints I will just include that check inside the Resource class. For now this can stay as it is.

wojtek0806:
Please fix the test issues before I can merge

wojtek0806:
Looks like both Travis and Jenkins are suffering from the bug in requests, i will provisionally merge this, we we are not accepting any new PRs until we fix the CI

PR #1470, opened 2018-06-26
Implement REST proxy for BIG-IQ

buzz-tee:
The PR adapts some of the strategies from the iWorkflow REST proxy implementation and allows using the REST proxy on BIG-IQ. The "isRestProxyEnabled" flag will be set, if required (in contrast to iWorkflow it is ***not*** set automatically by BIG-IQ during discovery).

Example:
```python
from bigiq import ManagementRoot
...
bigip = ManagementRoot(bigiq_hostname, username, password, proxy_to=bigip_hostname)
```

To skip enabling the REST proxy, simply pass in an additional parameter: [`enable_proxy=False`]

It also implement an option to pass in an existing iControlRestSession. This is useful when running as a server side application where it may be advisable to reuse an existing session in order to prevent re-authentication every other request.

Example (using Django's caching facilities):
```python
def get_bigiq(hostname, username, password, **kwargs):
    data = { 
        'hostname' : hostname, 
        'username' : username, 
        'password' : password,
    }

    key = 'bigiq-{}'.format(sha1(str(sorted(data.items()))).hexdigest())
    icrs = cache.get(key, None)
    if icrs:
        # Cache hit
        kwargs['icrs'] = icrs
        bigiq = ManagementRoot(hostname, username, password, **kwargs)
    else:
        # Cache miss
        bigiq = ManagementRoot(hostname, username, password, **kwargs)
        icrs = bigiq.icrs
    
    cache.set(key, icrs, getattr(settings, 'CACHE_ICRS_MAXAGE', 600))
    return bigiq
```

When running a server side application, e.g. via WSGI it may be advisable to not register any signal handlers, see e9507f7. Disable it using the following line in your wsgi.py
```python
os.environ.setdefault('F5_NO_SIGNAL', '1')
```

### DIFF
--- a/f5/bigip/__init__.py
+++ b/f5/bigip/__init__.py
@@ -23,10 +23,15 @@ try:
 except ImportError:
     import urllib.parse as urlparse
 
+from os import environ
+
 try:
-    import signal
-    from signal import SIGALRM
-    HAS_SIGNAL = True
+    if environ.get('F5_NO_SIGNAL') == '1':
+        HAS_SIGNAL = False
+    else:
+        import signal
+        from signal import SIGALRM
+        HAS_SIGNAL = True
 except ImportError:
     HAS_SIGNAL = False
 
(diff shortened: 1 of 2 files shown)


buzz-tee:
No idea why build check is failing, maybe somebody with access to f5net.com can check?
I'm using the code as drop-in replacement for the original sdk classes with no issues.
Thanks

wojtek0806:
@caphrim007 can you look at this if you have time, as you were writing iWorkflow and BigIQ endpoints for this

PR #1471, opened 2018-07-06
added firewall rule creation under policy

ujwalbendale:
Problem : In current implementation we can create firewall rules under rules list, but the path to attach rule lists to policy is broken, there is no rule-list sub-collection under policy collection, so it is better way to create rules under policy itself because rule-lists cannot be attached to policy.

Solution: Changed rule creation required kind to policy --> rules (tm:security:firewall:policy:rules:rulesstate)
so we can create rule under policy instead of rule_list (tm:security:firewall:rule-list:rules:rulesstate)

files changed- 
f5/bigip/tm/security/firewall.py
f5/bigip/tm/security/test/functional/test_firewall.py
f5/bigip/tm/security/test/unit/test_firewall.py

### DIFF
--- a/f5/bigip/tm/security/firewall.py
+++ b/f5/bigip/tm/security/firewall.py
@@ -124,13 +124,13 @@ class Rule_List(Resource):
 
 class Rules_s(Collection):
     """BIG-IP® AFM® Rules sub-collection."""
-    def __init__(self, rule_list):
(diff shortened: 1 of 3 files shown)


nberezin:
@caphrim007 @jasonrahm can you please review the PR? This change is critical for us.

PR #1501, opened 2018-08-11
Correct edge case in ltm node api for address + fqdn

### LINKED ISSUE #1500: node API backwards compat break
The patch here https://github.com/F5Networks/f5-common-python/pull/1492/ appears to have broken a specific, valid, case with the node API where both `address` and `fqdn` are allowed so-long as the `address` value is `any6`.

This broke the bigip_node module in ansible

caphrim007:
Issues:
Fixes #1500

Problem:
The ltm node api has an edge case with the address value "any6" when
it is specified with the fqdn key.
The patch that originally fixed this issue was causing
the bigip_node module in ansible to break.

Analysis:
This patch handles that edge case and allows it.

Tests:
* functional

### DIFF
--- a/f5/bigip/tm/ltm/node.py
+++ b/f5/bigip/tm/ltm/node.py
@@ -62,6 +62,8 @@ class Node(Resource):
         has_any = [set(x).issubset(args) for x in required_one_of]
         if len([x for x in has_any if x is True]) == 1:
             return self._create(**kwargs)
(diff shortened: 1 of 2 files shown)


caphrim007:
@jasonrahm ready for review

jasonrahm (approved):
Lgtm

PR #1616, opened 2023-10-09
OPENSTACK-2894: use utf-8 code for bigip configuration

zhang-shengping:


zhang-shengping:
[I record the question at stackoverflow](https://stackoverflow.com/questions/77256177/how-python-request-method-set-its-json-parameter-with-a-utf-8-encoding-value)

zhang-shengping:
```python
# -*- coding: utf-8 -*-

from f5_openstack_agent.lbaasv2.drivers.bigip.resource \
    import VirtualServer
from f5.bigip import ManagementRoot
import json

partition = u'Project_6fd06a50b7824ae48386565786e94b38'
name = u'test'

bigip = ManagementRoot(
    "10.110.119.120",
    "admin",
    "xxxxx",
    timeout=10,
    debug=True
)

# this character, the modify PATCH will failed with bigip error.
test = u"退"

# if use this character, the modify PATCH will succeed.
# test = u"哈哈"

vs = VirtualServer()
server = vs.load(bigip, name=name, partition=partition)
server.modify(description=test)
```

The `test` is a unicode, when the python Request module send this unicode to bigip server, it uses `json` parameter, which dumps the request body as json bytestring, and the unicode is escaped. The bigip server cannot correctly decode the json bytestring.

the bigip GUI web displays unreadable messy code as below:

image


To resolve this problem, this patch encodes the request body into utf-8 bytestring and sets the bytestring to the request `data` parameter.

image

